Issue #76, opened 2012-09-23
documenting that jQuery doesn't work well when a page is zoomed


mikesherov:
Several tickets came up recently that were closed because we don't support zoom well. Not sure if those bugs would've been filed if we told the community that we don't support it:

http://bugs.jquery.com/ticket/12594
http://bugs.jquery.com/ticket/12590

Issue #103, opened 2012-10-15
Note in the dimension module methods that some browsers return floats instead of integers


mikesherov:
As noted in http://bugs.jquery.com/ticket/11401:

> Agreed. Some of the differences in browsers concerning decimal point values should be respected for the sake of animations, if nothing else. We should not hide these.

arthurvr:
This was fixed in https://github.com/jquery/api.jquery.com/commit/2123d42a43f5bb46b7564a81184cd224d7fb4b31, seems like GitHub didn't auto-close.

PR #434, opened 2014-01-27
Dimensions: values are non-integer and non-zoomy, fixes gh-76, gh-103

dmethvin:
This has been true for most older versions as well but we just didn't have the docs right.


arthurvr:
@dmethvin Could you delete the branch created for this PR? No need for it to exist anymore :)

Issue #240, opened 2013-01-28
Update api entries with 1.9 upgrade guide info


dmethvin:
This could be broken into separate issues if that's more convenient.
- [x] .toggle(function, function, ... ) removed
- [x] jQuery.browser() removed
- [x] .live() removed
- [x] .die() removed
- [x] jQuery.sub() removed
- [x] .add()
- [x] .addBack( selector ) replaces .andSelf()
- [ ] .after(), .before(), and .replaceWith() with disconnected nodes http://bugs.jquery.com/ticket/13338
- [x] .appendTo, .insertBefore, .insertAfter, and .replaceAll
- [x] AJAX events should be attached to document
- [x] Checkbox/radio state in a .trigger()ed "click" event
- [x] Order of triggered "focus" events
- [ ] jQuery(htmlString) versus jQuery(selectorString)
- [x] Events not fired by the .data() method; names with periods
- [ ] Ordering of disconnected nodes within a jQuery set
- [x] Loading and running scripts inside HTML content
- [x] .attr() versus .prop()
- [x] $("input").attr("type", newValue) in oldIE
- [x] "hover" pseudo-event
- [x] .selector property on jQuery objects
- [x] jQuery.attr()
- [x] jQuery.ajax returning a JSON result of an empty string
- [x] jQuery.proxy() context
- [x] .data("events")
- [x] Removed properties of the Event object

PR #436, opened 2014-01-27
Note new sane 1.9+ behavior for disconnected elems, ref gh-240

dmethvin:
Since behavior of older versions was flakey with disconnected elements I decided to just remove the reference rather than explain it, since people should avoid the practice with these APIs anyway


arthurvr:
@dmethvin Could you delete the branch created for this PR? No reason for it to exist anymore :)

PR #453, opened 2014-03-07
The docs for $.Callbacks().has indicate that an argument is required

linusthe3rd:
The documentation for $.Callbacks().has does not describe that the argument for the method is actually optional even though the source code for the method says otherwise[1].

I tried to add a suitable description to the method to indicate this. If it is not sufficient enough, I will go ahead and open an issue to update the docs instead.

[1] https://github.com/jquery/jquery/blob/master/src/callbacks.js#L145

### DIFF
--- a/entries/callbacks.has.xml
+++ b/entries/callbacks.has.xml
@@ -3,11 +3,11 @@
   <title>callbacks.has()</title>
   <signature>
     <added>1.7</added>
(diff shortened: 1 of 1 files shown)


arthurvr:
For the reference, this fixed https://github.com/jquery/api.jquery.com/issues/549.

PR #511, opened 2014-06-14
Added a standards divergence warning to binders

phistuck:
Returning false in jQuery event handlers stops the propagation of the event, while returning false in native event listeners only prevents the default.
This adds a bold note to some bind methods (part I).


phistuck:
I will probably not get to it. Anyone of you is free to take this. Closing this.

PR #528, opened 2014-07-13
Update css.xml

AurelioDeRosa:
Added a sentence to make it clear that passing `null` as the second parameter of `css()` can be used to cancel any style modification previously performed.


AurelioDeRosa:
@gnarf I fixed the typo. In regard of the message, I guess it was a mistake because what I meant was "an empty string" and not `null`. In fact I didn't add anything about `null` as you noted. However, the text I added is valid.

AurelioDeRosa:
Can we merge this PR @gnarf? Are you happy with the change?

arthurvr:
LGTM. Merging this evening.

PR #533, opened 2014-07-17
Update attr.xml

AurelioDeRosa:
Removed incorrect sentence (possibly something that was valid in older versions of jQuery?) as proved here: http://jsfiddle.net/73cKp/1/


AurelioDeRosa:
Hi @gnarf. I've updated the PR based on the discussion. I don't why it wasn't already changed, but I guess I posted the message in the wrong thread.

arthurvr:
Thanks, @AurelioDeRosa! Does this PR look good now, @gnarf @dmethvin?

AurelioDeRosa:
@arthurvr No objections in a month, I think you can merge this PR.

PR #562, opened 2014-09-18
Clarified confusion for ':checked' selector regarding select inputs.

### LINKED ISSUE #558: :checked selector does not work for select elements
On the ":checked Selector" page (http://api.jquery.com/checked-selector/), the description states:

"The :checked selector works for checkboxes, radio buttons, **and select elements**. For select elements only, use the :selected selector."

The :checked selector does not work on select elements. It works on the `option` children of select elements (see http://jsfiddle.net/jhfrench/mr7koj73/ for demonstration).

Suggest enumerating the destinction with this description:
"The :checked selector works for checkboxes, radio buttons, **and "selected" options of select elements**. For **the selected options of** select elements only, use the :selected selector."

I think it would also be a good idea to include a third example which illustrates using `:checked` with select/option elements.

mcmathews:
Closes #558

### DIFF
--- a/entries/checked-selector.xml
+++ b/entries/checked-selector.xml
@@ -7,7 +7,7 @@
   </signature>
   <desc>Matches all elements that are checked or selected.</desc>
   <longdesc>
-    <p>The <code>:checked</code> selector works for checkboxes, radio buttons, and select elements. For select elements only, use the <code>:selected</code> selector.</p>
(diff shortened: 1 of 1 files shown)


arthurvr:
ping @mrclay & @mcmathews - One of you interested in updating the PR or creating a new?

mrclay:
Replaced by #616

PR #592, opened 2014-11-17
Update position.xml

pcj:
Clarify documentation to explicitly specify position() is not supported on text elements.

### DIFF
--- a/entries/position.xml
+++ b/entries/position.xml
@@ -9,7 +9,7 @@
     <p>The <code>.position()</code> method allows us to retrieve the current position of an element <em>relative to the offset parent</em>. Contrast this with <code><a href="/offset/">.offset()</a></code>, which retrieves the current position <em>relative to the document</em>. When positioning a new element near another one and within the same containing DOM element, <code>.position()</code> is the more useful.</p>
     <p>Returns an object containing the properties <code>top</code> and <code>left</code>.</p>
     <div class="warning">
-      <p><strong>Note:</strong> jQuery does not support getting the position coordinates of hidden elements or accounting for borders, margins, or padding set on the body element.</p>
+      <p><strong>Note:</strong> jQuery does not support getting the position coordinates of hidden elements or accounting for borders, margins, or padding set on the body element.  Position coordinates are not supported for node types other than Element (such as text nodes).</p>
     </div>
   </longdesc>
 	<note id="dimensions-number" type="additional" data-title=".position()"/>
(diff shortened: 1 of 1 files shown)


AurelioDeRosa:
Is this PR good to be merged? On a side note, I think @dmethvin note could be added. If anyone agrees I can submit a PR.

entries/position.xml
```diff
@@ -9,7 +9,7 @@
     <p>The <code>.position()</code> method allows us to retrieve the current position of an element <em>relative to the offset parent</em>. Contrast this with <code><a href="/offset/">.offset()</a></code>, which retrieves the current position <em>relative to the document</em>. When positioning a new element near another one and within the same containing DOM element, <code>.position()</code> is the more useful.</p>
     <p>Returns an object containing the properties <code>top</code> and <code>left</code>.</p>
     <div class="warning">
-      <p><strong>Note:</strong> jQuery does not support getting the position coordinates of hidden elements or accounting for borders, margins, or padding set on the body element.</p>
+      <p><strong>Note:</strong> jQuery does not support getting the position coordinates of hidden elements or accounting for borders, margins, or padding set on the body element.  Position coordinates are not supported for node types other than Element.</p>
```
arthurvr: There are two spaces before the word `Position`.

arthurvr:
> The default assumption should be that jQuery operations on text nodes are not supported. They're not supported by .attr(), .css(), .offset(), .position() .addClass(), .find(), any of the event APIs, and many others.

Do we really need a note on all these page (as seen in this PR)? I don't think so. IMHO a paragraph on the "using dom elements" article is enough:

> Ah, how about if we clarify that the jQuery(element) signature is not very useful for wrapping text nodes? Probably a paragraph in the "Using DOM Elements" section. I'm assuming you got a text node into a jQuery object that way.

If everybody agrees with that, we should open a decent ticket for that page and close this PR.

arthurvr:
> If everybody agrees with that, we should open a decent ticket for that page and close this PR.

@dmethvin @AurelioDeRosa Thoughts? I don't think having an explicit note on many of our pages makes sense, if the default assumption should be it isn't supported.

dmethvin:
Agreed, @arthurvr. Let's improve the messaging on `jQuery()` and perhaps [.contents()](http://api.jquery.com/contents/) which is another place where text nodes can sneak in. 

We never got a reply from @pcj on how the text node ended up in the collection, but if we do we could try to address that usage as well.

arthurvr:
> Agreed, @arthurvr. Let's improve the messaging on jQuery() and perhaps .contents() which is another place where text nodes can sneak in.

There we go: https://github.com/jquery/api.jquery.com/issues/709

Issue #345, opened 2013-07-23
Note that JS-initiated change does not trigger change event


dmethvin:
http://api.jquery.com/change/
http://api.jquery.com/val/
http://bugs.jquery.com/ticket/14171

PR #617, opened 2015-01-03
change: note that JS-initiated change does not trigger change event

arthurvr:
This would fix https://github.com/jquery/api.jquery.com/issues/345.

### DIFF
--- a/entries/change.xml
+++ b/entries/change.xml
@@ -52,6 +52,9 @@ $( "#other" ).click(function() {
     </code></pre>
     <p>After this code executes, clicks on <samp>Trigger the handler</samp> will also alert the message. The message will display twice, because the handler has been bound to the <code>change</code> event on both of the form elements.</p>
     <p>As of jQuery 1.4, the <code>change</code> event bubbles in Internet Explorer, behaving consistently with the event in other modern browsers.</p>
+    <div class="warning">
+	  <p><strong>Note: </strong>changing the value of an input element using JavaScript, using <a href="/val"><code>.val()</code></a> for example, won't fire the event.</p>
+    </div>
   </longdesc>
   <example>
     <desc>Attaches a change event to the select that gets the text for each selected option and writes them in the div.  It then triggers the event for the initial text draw.</desc>


entries/change.xml
```diff
@@ -52,6 +52,9 @@ $( "#other" ).click(function() {
     </code></pre>
     <p>After this code executes, clicks on <samp>Trigger the handler</samp> will also alert the message. The message will display twice, because the handler has been bound to the <code>change</code> event on both of the form elements.</p>
     <p>As of jQuery 1.4, the <code>change</code> event bubbles in Internet Explorer, behaving consistently with the event in other modern browsers.</p>
+    <div class="warning">
+      <p><strong>Note: </strong> dynamically changing the value of an input element, using <a href="/val"><code>.val()</code></a> for example, won't fire the event.</p>
```
dmethvin: The word "dynamically" is a little bit ambiguous here, since someone might interpret a user-generated change to be dynamic. Perhaps "Changing the value property of an input element using JavaScript..." instead?

entries/val.xml
```diff
@@ -130,6 +130,9 @@ $( "input:text.items" ).val(function( index, value ) {
 });
       </code></pre>
       <p>This example appends the string " items" to the text inputs' values.</p>
+        <div class="warning">
+          <p><strong>Note: </strong> changing the value of an input element using this method won't fire the <code><a href="/change">change</a></code> event.</p>
```
dmethvin: Do we need this in the `.val()` method as well? Lots of copy/pasted fragments make for a mess when it comes to updates. And yes we can turn these into reusable chunks but I'm questioning whether it's needed here since someone should be looking to event docs if they're listening for an event.

arthurvr: I agree with the point that having the same note on multiple places isn't maintainable.

> whether it's needed here since someone should be looking to event docs if they're listening for an event

I do think a note can't hurt. People will be at least on one of these pages when something goes wrong. In fact the [original ticket for this](http://bugs.jquery.com/ticket/14171) proves that, it had the title _".val(var) does not trigger .change"_

entries/change.xml
arthurvr: Agreed. I'il update my PR this evening.

arthurvr:
@dmethvin Does this LGTY now? I did the updates you asked for.

dmethvin:
I think the "changing" should be "Changing", otherwise good.

PR #625, opened 2015-01-15
Refer to jQuery.Deferred as a factory rather than a constructor.

ChadKillingsworth:
The `jQuery.Deferred` description describes it as a constructor. This is misleading. The returned object fails the following checks:

```JavaScript
(new jQuery.Deferred()) instanceof jQuery.Deferred === true //this is false
jQuery.Deferred() instanceof jQuery.Deferred === true //this is false
```

Also, extending the prototype (inheriting) from the jQuery.Deferred method would not produce the expected results. (it would infact do nothing useful).

This update describes the method as a factory which returns a new `deferred` object which is consistent with the rest of the api site documentation.

For more info, see google/closure-compiler#767

I should have a CLA on file - if not, let me know.

### DIFF
--- a/entries/jQuery.Deferred.xml
+++ b/entries/jQuery.Deferred.xml
@@ -14,10 +14,10 @@
       </argument>
     </argument>
   </signature>
-  <desc> A constructor function that returns a chainable utility object with methods to register multiple callbacks into callback queues, invoke callback queues, and relay the success or failure state of any synchronous or asynchronous function.</desc>
+  <desc> A factory function that returns a chainable utility object with methods to register multiple callbacks into callback queues, invoke callback queues, and relay the success or failure state of any synchronous or asynchronous function.</desc>
   <longdesc>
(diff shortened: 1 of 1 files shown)


arthurvr:
> I should have a CLA on file - if not, let me know.

For me it seems like you don't. Remember that the email and name in your git config must match the ones you signed the CLA with. If you already signed it with other name/email, you can change your git config or re-sign the CLA.

ChadKillingsworth:
You are correct - wrong email (just switched my global config recently). Author information changed - and new CLA signed.

arthurvr:
@dmethvin @timmywil Are we okay with this change?

timmywil:
Factory is more accurate, but it may not always be a factory. Regardless, it's probably better than referring to it as a constructor.

arthurvr:
> Factory is more accurate, but it may not always be a factory. Regardless, it's probably better than referring to it as a constructor.

Shall I land this, or do we want to think about another name for it?

timmywil:
Sorry to confuse. You should land.

PR #627, opened 2015-01-18
Submit-selector: Fix indentation for closing style tag

agcolom:


arthurvr:
Yeah, I believe this is the right thing, but doesn't the rest of the indentation on that page need to be changed then too?

agcolom:
I don't think so. Note that some of the indentation is taken care of by the xslt.

agcolom:
@arthurvr what other indentation do you see needs changing?

agcolom:
@arthurvr did you mean indent inside form?

arthurvr:
@agcolom Whoops! Looks good indeed, I was looking at the code above in split view, and that screwed it up.

Issue #413, opened 2014-01-09
Image error() handler example behaves incorrectly


theBRiGuy:
http://api.jquery.com/error/

The example's intention is to assign the new missing.png src ONLY if the image loads with an error.

Actual result: It seems that the example given behaves incorrectly. The image "missing.png" will be assigned to the selector's attribute regardless of whether or not the image loaded with an error.

kswedberg:
@dmethvin : at which version were shortcut event methods deprecated?

dmethvin:
I'm pretty sure it was 1.9, but can't find the discussion.

dmethvin:
Oh but it was just `.load()` and `.error()` which are strange and ambiguous.

kswedberg:
ah, ok. thanks for that info!

PR #630, opened 2015-01-22
Update error.xml

AurelioDeRosa:
Fix issue #413.

### DIFF
--- a/entries/error.xml
+++ b/entries/error.xml
@@ -52,7 +52,7 @@ $( "img" )
   })
   .attr( "src", "missing.png" );
 ]]></code>
-    <p>In the example, missing.png is an hypotetical missing image that is used to trigger the <code>error</code> event, not an image to replace those giving an error. To do that, you have to modify the body of the anonymous function passed to <code>.error()</code> as shown below:</p>
+    <p>In the example, missing.png is an hypothetical missing image that is used to trigger the <code>error</code> event, not an image to replace those giving an error. To do that, you have to modify the body of the anonymous function passed to <code>.error()</code> as shown below:</p>
     <code><![CDATA[
 $( "img" )
   .error(function() {
--- a/entries/error.xml
+++ b/entries/error.xml
@@ -20,7 +20,7 @@
     </argument>
   </signature>
   <longdesc>
-    <p>This method is a shortcut for <code>.on( "error", handler )</code>.</p>
+    <p><b>As of jQuery 1.8</b>, the <code>.error()</code> method is deprecated. Use <code>.on( "error", handler )</code> to attach event handlers to the <code>error</code> event instead.</p>
     <p>The <code>error</code> event is sent to elements, such as images, that are referenced by a document and loaded by the browser. It is called if the element was not loaded correctly.</p>
     <p>For example, consider a page with a simple image element:</p>
     <pre><code>
@@ -34,17 +34,17 @@ $( "#book" )
   })
   .attr( "src", "missing.png" );
     </code></pre>
-    <p>If the image cannot be loaded (for example, because it is not present at the supplied URL), the alert  is displayed:</p>
+    <p>If the image cannot be loaded (for example, because it is not present at the supplied URL), the alert is displayed:</p>
     <p>
       <samp>Handler for .error() called.</samp>
     </p>
     <div class="warning">
-      <p>The event handler <em>must</em> be attached before the browser fires the error event, which is why the example sets the src attribute after attaching the handler. Also, the error event may not be correctly fired when the page is served locally; <code>error</code> relies on HTTP status codes and will generally not be triggered if the URL uses the <code>file:</code> protocol.</p>
+      <p>The event handler <em>must</em> be attached before the browser fires the <code>error</code> event, which is why the example sets the <code>src</code> attribute after attaching the handler. Also, the <code>error</code> event may not be correctly fired when the page is served locally; <code>error</code> relies on HTTP status codes and will generally not be triggered if the URL uses the <code>file:</code> protocol.</p>
     </div>
-    <p>Note: A jQuery error event handler should not be attached to the window object. The browser fires the window's error event when a script error occurs. However, the window error event receives different arguments and has different return value requirements than conventional event handlers. Use <code>window.onerror</code> instead.</p>
+    <p>Note: A jQuery <code>error</code> event handler should not be attached to the <code>window</code> object. The browser fires the <code>window</code>'s <code>error</code> event when a script error occurs. However, the window <code>error</code> event receives different arguments and has different return value requirements than conventional event handlers. Use <code>window.onerror</code> instead.</p>
   </longdesc>
   <example>
-    <desc>To hide the "broken image" icons for IE users, you can try:</desc>
+    <desc>To hide the "broken image" icons for users of older versions of IE, you can write:</desc>
     <code><![CDATA[
 $( "img" )
   .error(function() {
@@ -52,6 +52,15 @@ $( "img" )
   })
   .attr( "src", "missing.png" );
 ]]></code>
+    <p>In the example, missing.png is an hypotetical missing image that is used to trigger the <code>error</code> event, not an image to replace those giving an error. To do that, you have to modify the body of the anonymous function passed to <code>.error()</code> as shown below:</p>
+    <code><![CDATA[
+$( "img" )
+  .error(function() {
+    $( this ).attr( "src", "replacement.png" );
+  })
+  .attr( "src", "missing.png" );
+]]></code>
+  <p><b>Note:</b> Be sure that the replacement image exists otherwise the <code>error</code> event will be triggered indefinitely.</p>
   </example>
   <category slug="events/browser-events"/>
   <category slug="version/1.0"/>
--- a/entries/error.xml
+++ b/entries/error.xml
@@ -20,6 +20,7 @@
     </argument>
   </signature>
   <longdesc>
+    <p>This method is a shortcut for <code>.on( "error", handler )</code>.</p>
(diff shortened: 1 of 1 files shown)


entries/error.xml
```diff
@@ -52,6 +52,15 @@ $( "img" )
   })
   .attr( "src", "missing.png" );
 ]]></code>
+    <p>In the example, missing.png is an hypotetical missing image that is used to trigger the <code>error</code> event, not an image to replace those giving an error. To do that, you have to modify the body of the anonymous function passed to <code>.error()</code> as shown below:</p>
```
dmethvin: hypothetical

entries/error.xml
```diff
@@ -52,6 +52,15 @@ $( "img" )
...
+    <code><![CDATA[
+$( "img" )
+  .error(function() {
+    $( this ).attr( "src", "replacement.png" );
+  })
+  .attr( "src", "missing.png" );
+]]></code>
+  <p><b>Note:</b> Be sure that the replacement image exists otherwise the <code>error</code> event will be triggered indefinitely.</p>
```
dmethvin: Good note!

entries/error.xml
AurelioDeRosa: Typo corrected. Sorry about that.

entries/error.xml
kswedberg: Please add a semicolon after "exists"

AurelioDeRosa: Done :)

arthurvr:
This PR looks good to me. Shall I merge?

entries/error.xml
```diff
@@ -52,6 +52,15 @@ $( "img" )
...
 ]]></code>
+    <p>In the example, missing.png is an hypothetical missing image that is used to trigger the <code>error</code> event, not an image to replace those giving an error. To do that, you have to modify the body of the anonymous function passed to <code>.error()</code> as shown below:</p>
+    <code><![CDATA[
+$( "img" )
+  .error(function() {
+    $( this ).attr( "src", "replacement.png" );
+  })
+  .attr( "src", "missing.png" );
+]]></code>
+  <p><b>Note:</b> Be sure that the replacement image exists; otherwise the <code>error</code> event will be triggered indefinitely.</p>
```
arthurvr: the indentation for this line seems wrong.

arthurvr: Also, I don't think `<p>` tags are allowed in `<example>`s. This just results into this: http://gyazo.com/f7795ca0fd998b6701b68bbb322fed0d (notice your extra info inside the `<p>` isn't rendered).

@AurelioDeRosa Do you have a working local docs setup? Normally you should be able to repro that there. If you don't, ping me on IRC and I'll  be happy to help you out.

arthurvr: ping @AurelioDeRosa

AurelioDeRosa: @arthurvr I'm going to take care of this issue right now. Sorry for the delay.

AurelioDeRosa:
As @arthurvr noted, the current version of the XSL file doesn't allow to have a paragraph in an `example`. So, the only solution I see is to allow it. Who can fix this issue? Any alternatives? // cc @dmethvin @kswedberg @scottgonzalez

dmethvin:
I think that notes were intended for small notes and not big clips with paragraphs and examples. If we allow big clips and block elements like code, it may affect layout in a drastic way so that's something to take into consideration.

scottgonzalez:
Well you for sure can't just throw arbitrary markup into `<example>`. We can potentially add support for `<longdesc>` inside `<example>`, but the currently used markup will never be allowed.

AurelioDeRosa:
What's your suggestion to rearrange the code to have a description?

scottgonzalez:
That's hard to say because I can't tell what you're trying to do. Are you trying to create an inline runnable example that is accompanied by a code block explaining how to do something? This is very different from the way any other page is designed.

arthurvr:
In this case I'd opt to move some stuff in the `<desc>` and other things can fit in a JavaScript comment.

> We can potentially add support for <longdesc> inside <example>, but the currently used markup will never be allowed.

This is kinda off-topic for this PR, and should be discussed in `grunt-jquery-content` isn't it?

scottgonzalez:
> In this case I'd opt to move some stuff in the <desc> and other things can fit in a JavaScript comment.

The `<desc>` is really a title. So I don't think that's an acceptable solution.

> >  We can potentially add support for inside , but the currently used markup will never be allowed.

> This is kinda off-topic for this PR, and should be discussed in grunt-jquery-content isn't it?

We certainly shouldn't open an issue for grunt-jquery-content until we know what we're actually discussing. I'd like to hear from @AurelioDeRosa since he sent the PR which contains two `<code>` blocks inside `<example>`, along with other HTML content.

AurelioDeRosa:
@scottgonzalez The PR addresses #413. The idea is to have a description of what's going on in the example for the benefit of some users.

scottgonzalez:
Sounds like what you really want is a new, useful example.

AurelioDeRosa:
I'd not say that a new example will solve the problem. The problem was the user didn't get what the example really did. Perhaps @arthurvr's suggestion of having a few comments in the code can fix this.

AurelioDeRosa:
@arthurvr @scottgonzalez Should I try to submit a PR which has JavaScript comment so that we can review it?

scottgonzalez:
sure

arthurvr:
I'd love to move forward on this PR. ping @AurelioDeRosa

AurelioDeRosa:
PR updated @arthurvr.

entries/error.xml
```diff
@@ -20,7 +20,7 @@
     </argument>
   </signature>
   <longdesc>
-    <p>This method is a shortcut for <code>.on( "error", handler )</code>.</p>
+    <p><b>As of jQuery 1.8</b>, the <code>.error()</code> method is deprecated. Use <code>.on( "error", handler )</code> to attach event handlers to the <code>error</code> event instead.</p>
```
arthurvr: This is a totally unrelated change. Mind extracting this to its own commit?

arthurvr: ping @AurelioDeRosa ;)

arthurvr: hey-hey @AurelioDeRosa still interested in picking this up?

arthurvr: ping @AurelioDeRosa

AurelioDeRosa:
PR updated @arthurvr. Sorry for the late change.

arthurvr:
@AurelioDeRosa I liked the change, it was just unrelated to the thing you're trying to fix, so it should have been in a separated commit. I didn't mean you should remove the change :wink:

AurelioDeRosa:
@arthurvr I was thinking about opening a different PR for it, that's why I removed it.

arthurvr:
Makes sense. Thanks.

entries/error.xml
```diff
@@ -20,7 +20,6 @@
     </argument>
   </signature>
   <longdesc>
-    <p>This method is a shortcut for <code>.on( "error", handler )</code>.</p>
```
arthurvr: This paragraph shouldn't be removed (or at least not in this commit).

AurelioDeRosa: Oh... 🐑, you're right!

AurelioDeRosa:
Can someone check the updated PR so that we can merge it?

entries/error.xml
```diff
@@ -34,18 +34,19 @@ $( "#book" )
...
     </div>
-    <p>Note: A jQuery error event handler should not be attached to the window object. The browser fires the window's error event when a script error occurs. However, the window error event receives different arguments and has different return value requirements than conventional event handlers. Use <code>window.onerror</code> instead.</p>
+    <p>Note: A jQuery <code>error</code> event handler should not be attached to the <code>window</code> object. The browser fires the <code>window</code>'s <code>error</code> event when a script error occurs. However, the window <code>error</code> event receives different arguments and has different return value requirements than conventional event handlers. Use <code>window.onerror</code> instead.</p>
   </longdesc>
   <example>
-    <desc>To hide the "broken image" icons for IE users, you can try:</desc>
+    <desc>To hide the "broken image" icons for users of older versions of IE, you can write:</desc>
```
arthurvr: > of older versions of IE

Which versions does that affect?

AurelioDeRosa: I remember is IE <= 10 but I can't find any reference. Someone with a hint on this?

arthurvr: No idea, honestly. I do think though that just saying "older versions of IE" means nothing so we should figure this out. Maybe @dmethvin or @timmywil have links to relevant discussions about this?

arthurvr: ping @AurelioDeRosa @dmethvin @timmywil is there anyone who knows this? If not I'll quickly test this out.

dmethvin: Yeah, I mentioned elsewhere that I'm not sure when the "broken image" thing went away. Probably before IE9?

AurelioDeRosa: Should we specify IE9 for now and update in case of mistake?

AurelioDeRosa:
This PR is standing from a few months. Is anyone happy to specify IE9 and merge it?

dmethvin:
Looks like this has been settable from at least IE and Firefox for years:

http://stackoverflow.com/questions/1694937/firefox-show-broken-images
http://ccm.net/faq/13546-internet-explorer-show-image-download-placeholders

However, I couldn't find out when the default was changed. How about if we just delete the example? There is still a good one there.

AurelioDeRosa:
@dmethvin Thank you for the feedback. I've updated the PR based on your comment.

arthurvr:
lgtm :+1:

Issue #384, opened 2013-11-11
.off(Event_object) isn't documented


dmethvin:
It looks like the entry from http://api.jquery.com/unbind/#unbind-event didn't get copied when `.off()` was created.

AurelioDeRosa:
I've sent a PR (actually two, but please ignore the first) to add this info.

On a related note, I can see that the documentation of `unbind()` specifies the argument as <a href="http://api.jquery.com/Types/#Object"><code>Object</code></a>, but it should be <a href="http://api.jquery.com/category/events/event-object/"><code>jQuery.Event</code></a>, shouldn't it?

dmethvin:
Yes, that should be `jQuery.Event` as well for `.unbind()`.

PR #634, opened 2015-01-23
Update off.xml

AurelioDeRosa:
Fix issue #384.

### DIFF
--- a/entries/off.xml
+++ b/entries/off.xml
@@ -26,6 +26,12 @@
   </signature>
   <signature>
     <added>1.7</added>
+    <argument name="event" type="Event">
+      <desc>A <a href="/category/events/event-object/"><code>jQuery.Event</code></a> object.</desc>
+    </argument>
+  </signature>
+  <signature>
+    <added>1.7</added>
   </signature>
   <longdesc>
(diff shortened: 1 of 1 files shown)


arthurvr:
This PR adds a `<signature>`, but I'd like to see a little more. A more in-depth explanation, an example maybe?

AurelioDeRosa:
@arthurvr I didn't include an example because the `one()` method has one that uses this signature and I didn't want to replicate it.

entries/off.xml
```diff
@@ -26,6 +26,12 @@
   </signature>
   <signature>
     <added>1.7</added>
+    <argument name="event" type="Event">
+      <desc>A <a href="/category/events/event-object/"><code>jQuery.Event</code></a> object.</desc>
+    </argument>
+  </signature>
+  <signature>
+    <added>1.7</added>
```
dmethvin: This signature was around earlier than 1.7, I had to keep it working when I refactored the code in 1.7. I'll look for the right version.

dmethvin: Oh but not in `.off()` :smile: Carry on.

arthurvr:
LGTM. Shall I merge?

kswedberg:
@arthurvr :+1:

Issue #636, opened 2015-01-26
In index html escape the descriptions 


rasca:
When visiting http://api.jquery.com/ in jQuery.param() description it should say:

> Create a serialized representation of an array, a plain object, or a jQuery object suitable for use in a URL query string or Ajax request. In case a jQuery object is passed, it should contain \<input\> elements with name/value properties.

But instead of `<input>` it displays an input.

AurelioDeRosa:
Thank you for catching this issue @rasca.

rasca:
@AurelioDeRosa thanks for the quick fix! Maybe it can be set to escape the xml text when rendering so this won't happen again in the future?

AurelioDeRosa:
@rasca I agree this would be useful (cc: @arthurvr @kswedberg). In the meantime I fixed the entry to be inline with the other entries that don't have codes in the summary.

scottgonzalez:
If you're going to close this, then please open a new issue since the real problem still exists.

PR #637, opened 2015-01-26
Update jQuery.param.xml

AurelioDeRosa:
Fix issue #636

### DIFF
--- a/entries/jQuery.param.xml
+++ b/entries/jQuery.param.xml
@@ -22,7 +22,7 @@
       <desc>A Boolean indicating whether to perform a traditional "shallow" serialization.</desc>
     </argument>
   </signature>
-  <desc>Create a serialized representation of an array, a plain object, or a jQuery object suitable for use in a URL query string or Ajax request. In case a jQuery object is passed, it should contain <code>&lt;input&gt;</code> elements with name/value properties.</desc>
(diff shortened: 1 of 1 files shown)


scottgonzalez:
This doesn't seem like a legit fix.

AurelioDeRosa:
Why? If you refer to the fact that I removed the code is because all the other entries don't have code in the summary. Take as an example ":input Selector" and ":text Selector".

scottgonzalez:
So there are other places that *should* use `<code>`. The bottom line is `<desc>` supports markup and the index page should be fixed.

AurelioDeRosa:
Cool. That's exactly what I asked [here](https://github.com/jquery/api.jquery.com/issues/636#issuecomment-71497916).

arthurvr:
> So there are other places that should use <code>. The bottom line is <desc> supports markup and the index page should be fixed.

@scottgonzalez Where should that be handled? Is it okay if I merge this in for now, and we open a separated issue for that?

scottgonzalez:
I'd prefer if it were properly fixed, since our yet to be properly written style guide will say that the current content is correct.

scottgonzalez:
With that being said, you can merge this if you want to.

arthurvr:
Merged and deployed. Thanks for being that quick with opening a PR, @AurelioDeRosa.

Issue #388, opened 2013-11-28
The unload docs have incorrect examples.


mzgol:
The page under http://api.jquery.com/unload/ shows an example with `alert` which no longer works (in Chrome, Firefox) due to browser changes. The only allowed thing to do in the `beforeunload` handler are returning a string.

Originally reported as a jQuery bug: http://bugs.jquery.com/ticket/14559

arthurvr:
@mzgol Just opened a PR: https://github.com/jquery/api.jquery.com/pull/639.

PR #639, opened 2015-01-28
Unload: update outdated example

arthurvr:
Would fix https://github.com/jquery/api.jquery.com/issues/388.

### DIFF
--- a/entries/unload.xml
+++ b/entries/unload.xml
@@ -31,9 +31,7 @@ $( window ).unload(function() {
   return "Handler for .unload() called.";
 });
     </code></pre>
-    <p>After this code executes, the alert will be displayed whenever the browser leaves the current page.
-It is not possible to cancel the <code>unload</code> event with <code>.preventDefault()</code>. This event is available so that scripts can perform cleanup when the user leaves the page.
-    </p>
+    <p>This event is available so that scripts can perform cleanup when the user leaves the page. Most browsers will ignore calls to <code>alert()</code>, <code>confirm()</code> and <code>prompt()</code> inside the event handler. The string you return may be used in a confirmation dialog, but not all browsers support this. It is not possible to cancel the <code>unload</code> event with <code>.preventDefault()</code>.</p>
   </longdesc>
   <example>
     <desc>To display an alert when a page is unloaded:</desc>
--- a/entries/unload.xml
+++ b/entries/unload.xml
@@ -28,7 +28,7 @@
     <p>Any <code>unload</code> event handler should be bound to the <code>window</code> object:</p>
     <pre><code>
 $( window ).unload(function() {
-  alert( "Handler for .unload() called." );
+  return "Handler for .unload() called.";
 });
     </code></pre>
     <p>After this code executes, the alert will be displayed whenever the browser leaves the current page.
@@ -39,7 +39,7 @@ It is not possible to cancel the <code>unload</code> event with <code>.preventDe
     <desc>To display an alert when a page is unloaded:</desc>
     <code><![CDATA[
 $( window ).unload(function() {
-  alert( "Bye now!" );
+  return "Bye now!";
 });
   ]]></code>
   </example>


mzgol:
I'd also add a note that in Firefox (and maybe in some other places?) a message is ignored.

arthurvr:
> I'd also add a note that in Firefox (and maybe in some other places?) a message is ignored.

I thought about it, but are jQuery API docs the right place to learn people JavaScript? This isn't a jQuery specific behaviour.

mzgol:
We're already making this event special by setting `returnValue` by ourselves; people using the event directly have to know that. Since we're taking some responsibility off users, I think we should inform them we're not fixing everything. :)

What do you think, @dmethvin?

entries/unload.xml
```diff
@@ -28,7 +28,7 @@
     <p>Any <code>unload</code> event handler should be bound to the <code>window</code> object:</p>
     <pre><code>
 $( window ).unload(function() {
-  alert( "Handler for .unload() called." );
+  return "Handler for .unload() called.";
 });
     </code></pre>
     <p>After this code executes, the alert will be displayed whenever the browser leaves the current page.
```
dmethvin: s/alert/message/ maybe?

s/will be/may be/ as well?

dmethvin:
I believe we've gotten bug reports about this in the past, so perhaps we can add something to the warning already there.

arthurvr:
Just for the reference, dropping in https://github.com/jquery/learn.jquery.com/issues/608 too. Might be worth it to mention it there.

arthurvr:
> I believe we've gotten bug reports about this in the past, so perhaps we can add something to the warning already there.

Added a sentence warning about it. I don't think we should go in-depth. @dmethvin What do you think?

entries/unload.xml
```diff
@@ -28,18 +28,17 @@
     <p>Any <code>unload</code> event handler should be bound to the <code>window</code> object:</p>
     <pre><code>
 $( window ).unload(function() {
-  alert( "Handler for .unload() called." );
+  return "Handler for .unload() called.";
 });
     </code></pre>
-    <p>After this code executes, the alert will be displayed whenever the browser leaves the current page.
-It is not possible to cancel the <code>unload</code> event with <code>.preventDefault()</code>. This event is available so that scripts can perform cleanup when the user leaves the page.
+	<p>Most browsers will ignore calls to <code>alert()</code>, <code>confirm()</code> and <code>prompt()</code> inside the event handler. Depending on the browser, the message you return will be used in a confirmation dialog. It is not possible to cancel the <code>unload</code> event with <code>.preventDefault()</code>. This event is available so that scripts can perform cleanup when the user leaves the page.
```
mzgol: A tab? ;)

arthurvr: whoops! Fixed!

arthurvr: @mzgol Does the rest LGTY?

mzgol:
What about the info that some browsers will ignore the message at all?

dmethvin:
I agree the comments here should be kept simple and not go too deep. It's always a bit frustrating when an API reference turns into a big "how to write JavaScript in the browser" article. Fortunately this page is pretty short at this point so we're not adding that much more to read.

>  This event is available so that scripts can perform cleanup when the user leaves the page.

This is a great point, maybe move this to the beginning in its own paragraph? That brings up the async operation trap, so perhaps add a sentence after it like, "Note that the browser may cancel or deny asynchronous operations attempted in this handler, such as AJAX or setTimeout."

> Depending on the browser, the message you return will be used in a confirmation dialog.

Maybe something like this instead? "The string you return may be used in a browser confirmation dialog, but not all browsers support this." The warning already covers this ground again so I don't think we need much more.

mzgol:
@dmethvin That sounds good. :)



-- 
Michał Gołębiowski

arthurvr:
@mzgol @dmethvin Updated PR.

dmethvin:
LGTM!

PR #641, opened 2015-02-02
Remove protocol-relative URLs

konklone:
Removes ([now-deprecated](http://www.paulirish.com/2010/the-protocol-relative-url/)) protocol-relative URLs in favor of https:// URLs.


entries2html.xsl
```diff
@@ -12,7 +12,7 @@
 	&lt;meta charset="utf-8"&gt;
 	&lt;title&gt;<xsl:value-of select="//entry/@name"/> demo&lt;/title&gt;<xsl:if test="css">
 	&lt;style&gt;<xsl:value-of select="css/text()"/>	&lt;/style&gt;</xsl:if>
-	&lt;script src="//code.jquery.com/jquery-1.10.2.js"&gt;&lt;/script&gt;<xsl:if test="code/@location='head'">
+	&lt;script src="https://code.jquery.com/jquery-1.10.2.js"&gt;&lt;/script&gt;<xsl:if test="code/@location='head'">
```
gnarf: Should we bump to 1.11.2 while we are at it?

arthurvr: Yes, same thing in the PR for the learn site.

AurelioDeRosa: Yes, I agree with @gnarf.

arthurvr: We discussed this in the meeting feb 11, and decided to do this separately. https://github.com/jquery/api.jquery.com/issues/651

Issue #326, opened 2013-06-14
Clarification of jQuery.fn.css method


markelog:
It might be better to clarify that jQuery.fn.css getter returns computed value but not "style property".

See http://bugs.jquery.com/ticket/14022

arthurvr:
@markelog I just opened #646.

PR #646, opened 2015-02-06
css: clarify getter returns computed value

arthurvr:
Would fix https://github.com/jquery/api.jquery.com/issues/326.


gnarf:
LGTM! :+1:

Issue #98, opened 2012-10-15
Document that innerWidth/OuterWidth are now setters as well


mikesherov:
usage:

`innerWidth(number)` will subtract the padding, in pixels from number, and set the css width of the element to the difference

`outerWidth(number)` will subtract the padding and border, in pixels from number, and set the css width of the element to the difference

`outerWidth(number, true)` will subtract the padding and border and margin, in pixels from number, and set the css width of the element to the difference

arthurvr:
outerHeight is the only one that doesn't have this signature yet.

PR #647, opened 2015-02-08
outerHeight: document the method as a setter

arthurvr:
Adds 2 new signatures to the `outerHeight` entry, see https://github.com/jquery/api.jquery.com/issues/98. Doesn't fix that issue yet since `outerWidth` needs to get done too.

@dmethvin This is nearly the same commit as https://github.com/jquery/api.jquery.com/commit/0051b225f73aae7b719afa2b8d2da2363a0c4082. Would you mind reviewing this?


gnarf:
Seems a good enough "starter" for those signatures :+1:

PR #648, opened 2015-02-09
outerWidth: document method as a setter

### LINKED ISSUE #98: Document that innerWidth/OuterWidth are now setters as well

arthurvr:
Nearly same PR as https://github.com/jquery/api.jquery.com/pull/647. Would finally fix #98.

### DIFF
--- a/entries/outerWidth.xml
+++ b/entries/outerWidth.xml
@@ -1,4 +1,6 @@
 <?xml version="1.0"?>
+<entries>
+  <desc>Get the current computed width for the first element in the set of matched elements, including padding and border.</desc>
 <entry type="method" name="outerWidth" return="Number">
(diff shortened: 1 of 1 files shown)


gnarf:
DO IT! DO IT NOW!

arnold-schwarzenegger-screaming-kindergarten-cop_480_poster

dmethvin:
Looks good to me as well! (Does that guy have a tumor?)

PR #654, opened 2015-02-16
toggle: renamed toggle(showOrHide) to toggle(doShow)

burka:
To make it clear that passing true shows and passing false hides. Otherwise if you take showOrHide at its word, it would do nothing if you pass false and would call the regular toggle() if you pass true ;-)

### DIFF
--- a/entries/toggle.xml
+++ b/entries/toggle.xml
@@ -24,8 +24,8 @@
   </signature>
   <signature>
     <added>1.3</added>
-    <argument name="showOrHide" type="Boolean">
-      <desc>A Boolean indicating whether to show or hide the elements.</desc>
+    <argument name="setShown" type="Boolean">
+      <desc>Use <code>true</code> to show the element or <code>false</code> to hide it.</desc>
     </argument>
   </signature>
 
@@ -75,13 +75,13 @@ $( "#clickme" ).click(function() {
     <p>The second version of the method accepts a Boolean parameter. If this parameter is <code>true</code>, then the matched elements are shown; if <code>false</code>, the elements are hidden. In essence, the statement:
     </p>
     <pre><code>
-$( "#foo" ).toggle( showOrHide );
+$( "#foo" ).toggle( setShown );
     </code></pre>
     <p>is equivalent to:</p>
     <pre><code>
-if ( showOrHide === true ) {
+if ( setShown === true ) {
   $( "#foo" ).show();
-} else if ( showOrHide === false ) {
+} else if ( setShown === false ) {
   $( "#foo" ).hide();
 }
     </code></pre>


entries/toggle.xml
```diff
@@ -24,7 +24,7 @@
   </signature>
   <signature>
     <added>1.3</added>
-    <argument name="showOrHide" type="Boolean">
+    <argument name="doShow" type="Boolean">
```
gnarf: Not sure that `doShow` does this any better...  How about `setShown` ?  

Also maybe add a description in the `<desc>` tag specifying `true` will show and `false` will hide?

gnarf:
Thanks for the submission! Sorry that my `showOrHide` name was confusing to :smile:

burka:
Glad if it clears things up. At least for me ;-) `setShown` or `doShow` both sound like the initiation of a action, both are valid to me and I think both would make it more clear. I have no definite opinion on that, so setShown is ok for me if you prefer it.

gnarf:
:+1: Looks great, thanks for the quick update!

AurelioDeRosa:
IMHO `doShow` doesn't clarify the concept. `doShow` seems tied to the show concept only, and not that passing `false`  hides the elements.

scottgonzalez:
How about `makeVisible` (or honestly, just leaving it as is).

gnarf:
I think the change to the description is important either way, what we call it doesn't matter that much to me, any name for that parameter is awkward.

AurelioDeRosa:
:+1: for leaving it as is

burka:
The description should be changed in any case. Currently if you read it you might make the right guess, but you can't infer the meaning of the parameter from the documentation without reading the code snippet further down.

I think all alternates are a little awkward, but faster to understand than `showOrHide`. `makeVisible` might be a little better, just `show` might be a little better, `doShow`, `setShown` are both better than the current version and could be used until someone has a even better idea.

gnarf:
I like it as you have it for now. Going to merge it.

kswedberg:
Too bad. I agree that the description is better, but `setShown` is worse. It implies that the method will either show it or it won't, when what it really does is show it or hide it depending on its current state (hence the "showOrHide").

AurelioDeRosa:
I agree with @kswedberg. It wasn't an urgent fix, so we could have discussed a bit more.

gnarf:
I know we went back and forth on it when documenting it originally I think
'showOrHide' was my own suggestion, and I'm seeing problems with it now.
'set shown' felt better.  Let's talk it over in an issue and come back to
it if needed?

gnarf:
If you have opinions on what we should call this, please go to #655 - Thanks @burka for the submission (and starting this discussion inadvertently)!

gnarf:
Also @AurelioDeRosa I wasn't ignoring your "I liked it the old way" comment before the merge, I just didn't read it until *after* I had merged :frowning: sorry

AurelioDeRosa:
Don't worry @gnarf, I wasn't complaining :)

PR #657, opened 2015-02-18
css: add note about `px` being the default unit

### LINKED ISSUE #656: Note auto-`px` for Number values in .css()
Something like this:

>  When a Number is passed as the `value`, jQuery will convert it to a String and add `px` to the end of that string. If the property requires units other than `px`, convert the value to a string and add the appropriate units before calling the method.

arthurvr:
Would fix #656


gnarf:
:+1:

Issue #658, opened 2015-02-18
Can we remove our silly JSON.stringify polyfill from examples?


gnarf:
http://caniuse.com/#search=json - support back to IE8

The example in `jQuery.extend()` has a pretty complex "fake stringify" method and seems overly complex for todays needs.  https://github.com/jquery/api.jquery.com/blob/03fcbfbd17b494cc1122b6727adfcd8101b0ea9b/entries/jQuery.extend.xml#L59

arthurvr:
I wouldn't do it in master yet, as people might just copy-paste and be confused because jQuery 1.x is compatible with IE6+. That being said, I think we can just do this in the `v3.0.0-docs` branch, as there we don't have to support IE8 anyways.

arthurvr:
I can PR if everybody agrees.

gnarf:
We could also just use http://api.jquery.com/jQuery.param/ instead?  I just
want to simply the example...

arthurvr:
JSON is a little more readable IMHO, but fine with that as well.

gnarf:
I Agree that JSON is indeed more readable... I just don't think that JSON
encoder we have in the example is at all readable, and it confuses the
examples quite a bit to try and parse it.

arthurvr:
PR gh-661.

PR #661, opened 2015-02-23
jQuery.extend: simplify way of logging objects

arthurvr:
In the past we had a little silly polyfill for `JSON.stringify` in the examples in the `jQuery.extend` entry. This overcomplicated the example so should be replaced by something simpler. We can perfectly use `JSON.stringify` directly once v3 released so opening this against the `v3.0.0-docs` branch.

Would fix gh-658.

### DIFF
--- a/entries/jQuery.extend.xml
+++ b/entries/jQuery.extend.xml
@@ -56,17 +56,8 @@ var object2 = {
 // Merge object2 into object1
 $.extend( object1, object2 );
 
-var printObj = typeof JSON !== "undefined" ? JSON.stringify : function( obj ) {
-  var arr = [];
-  $.each( obj, function( key, val ) {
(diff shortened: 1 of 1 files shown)


arthurvr:
@gnarf Would you mind giving this a review? :)

gnarf:
I'd be :+1: to take this into master, perhaps with a `// Assuming JSON.stringify - not available in IE<8` - so definitely :+1: here in v3 docs

arthurvr:
@gnarf Doing it with a comment seems fine to me. Than we can remove that comment once v3 releases, right?

arthurvr:
@gnarf Updated PR. Does this version LGTY for master?

gnarf:
Works for me

PR #669, opened 2015-02-28
Update README.md

AurelioDeRosa:
Added a note to troubleshoot a possible issue when using Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ To build and deploy your changes for previewing in a [`jquery-wp-content`](https
 
 The `xmllint` and `xsltproc` utilities need to be in your path. If you are on Windows, you can get libxml2 and libxslt from <a href="http://sourceforge.net/projects/gnuwin32/files/">GnuWin32</a>.
 
-**Note**: If you're using Windows and you receive the error "Error" when executing the task "build-xml-entries:all", try to add the DLL "libwinpthread-1.dll" in the root of the project.
+**Note**: If you're using Windows and you receive the error "Error" when executing the task `build-xml-entries:all`, try to add the DLL `libwinpthread-1.dll` in the root of the project.
 
 ## Style Guidelines
 
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ To build and deploy your changes for previewing in a [`jquery-wp-content`](https
 
 The `xmllint` and `xsltproc` utilities need to be in your path. If you are on Windows, you can get libxml2 and libxslt from <a href="http://sourceforge.net/projects/gnuwin32/files/">GnuWin32</a>.
 
+**Note**: If you're using Windows and you receive the error "Error" when executing the task "build-xml-entries:all", try to add the DLL "libwinpthread-1.dll" in the root of the project.
+
 ## Style Guidelines
 
 ### Prose Style & Grammar


scottgonzalez:
This would also need to be included in every API repo.

README.md
```diff
@@ -11,6 +11,8 @@ To build and deploy your changes for previewing in a [`jquery-wp-content`](https
 
 The `xmllint` and `xsltproc` utilities need to be in your path. If you are on Windows, you can get libxml2 and libxslt from <a href="http://sourceforge.net/projects/gnuwin32/files/">GnuWin32</a>.
 
+**Note**: If you're using Windows and you receive the error "Error" when executing the task "build-xml-entries:all", try to add the DLL "libwinpthread-1.dll" in the root of the project.
```
arthurvr: Please wrap the filenames and the grunt task in backticks.

AurelioDeRosa: Done.

AurelioDeRosa:
I can do that @scottgonzalez.

arthurvr:
We may want this file in the gitignore then too?

AurelioDeRosa:
Is this PR good to be merged @arschmitz?

arthurvr:
I would like somebody to actually confirm that file is needed/appropriate. I'm not using windows myself :)

README.md
```diff
@@ -11,6 +11,8 @@ To build and deploy your changes for previewing in a [`jquery-wp-content`](https
 
 The `xmllint` and `xsltproc` utilities need to be in your path. If you are on Windows, you can get libxml2 and libxslt from <a href="http://sourceforge.net/projects/gnuwin32/files/">GnuWin32</a>.
 
+**Note**: If you're using Windows and you receive the error "Error" when executing the task `build-xml-entries:all`, try to add the DLL `libwinpthread-1.dll` in the root of the project.
```
arthurvr: > you receive the error "Error" when executing the task

Is "error" the only thing it throws?

AurelioDeRosa: Yes, unfortunately.

arthurvr:
ping, somebody?

> I would like somebody to actually confirm that file is needed/appropriate. I'm not using windows myself :)

dmethvin:
If @AurelioDeRosa got everything working on Windows he's accomplished more than I ever could. Which version of Windows is this?

AurelioDeRosa:
I'm running on Windows 8.1

dmethvin:
I'm running Windows 8.1 as well and i can `grunt lint` or `grunt build-xml-entries:all` for api.jquery.com without errors. I am guessing that dll has already been installed on my system through some other app that needed it, but I have no idea which one. Do you use msysgit?

AurelioDeRosa:
@dmethvin I'm using http://git-scm.com/download/win

dmethvin:
Okay, must be something specific to that setup. I just looked and can't find that DLL anywhere on my drive, but that part of the build at least works fine.

dmethvin:
BTW, is that the recommended setup for Windows now? My quick search on contribute.jquery.org and jquery-wp-content didn't find anything. I've used cygwin and msysgit with success but never just Windows git by itself.

AurelioDeRosa:
I'm not sure what you mean by "recommended setup". Do you mean "Is this the coolest tool used on Windows today?" or "Is this the tool we, as the jQuery Foundation, suggest to people?".

dmethvin:
The latter. The README for core has links to git (the one you linked) and node for Windows, but I've always installed msysgit because, well, because it works.

AurelioDeRosa:
Yes, indeed the guidelines suggest the tool I'm using, but to be honest, I already had it installed and it's the one I've always used.

At the end of the day, I'd vote for adding the note. I doesn't have any drawbacks by its own and may help developers in solving a hard-to-spot issue. Also, it's very small. What do you think @dmethvin?

dmethvin:
It doesn't seem like this is a problem specific to api.jquery.com. Is there a reason why the api.jqueryui.com repo recommends a different source for the xml utils? Maybe the ones from there wouldn't throw the error.

scottgonzalez:
> Is there a reason why the api.jqueryui.com repo recommends a different source for the xml utils?

See #676.

arthurvr:
> See #676.

Yup, I will change those URLs in every repo after this one is merged.

gnarf:
This seems good to me, just bumping it.

Issue #665, opened 2015-02-24
.on() - add note about not being able to remove a listener during the event


timmywil:
From https://github.com/jquery/jquery/issues/2094

> My suggested resolution is just to add a sentence to the .on() docs clarifying that the handler list for an element is set when the event is first delivered.

AurelioDeRosa:
I don't have the privileges to assign this issue to myself, so I'm posting this message to let you know that I'm working on it.

arthurvr:
@AurelioDeRosa Thank you!

arthurvr:
PR #670

PR #670, opened 2015-02-28
Update on.xml

AurelioDeRosa:
Fixed issue #665


AurelioDeRosa:
@arthurvr Any thought on this PR?

dmethvin:
> It will not be executed the following times because it's correctly removed. 

This was confusing to me, maybe it was "the following times" since it isn't completely clear what that means. I think it could just be removed.

> This behavior goes against the <a href="http://www.w3.org/TR/DOM-Level-2-Events/events.html#Events-EventTarget-removeEventListener">W3C events specification</a>. It's worth noting that this is an edge case and this behavior allows to keep the logic simple.

I would just remove this. It's such an edge case that spending too much time explaining it is likely to confuse rather than enlighten people. The example helps for those who care.

AurelioDeRosa:
> This was confusing to me, maybe it was "the following times" since it isn't completely clear what that means. I think it could just be removed.

Are you against the note or the wording? I feel like this is an important information to have.

> It's worth noting that this is an edge case and this behavior allows to keep the logic simple.

Let's remove this part only?

dmethvin:
It's definitely a balance between too many words and not enough. How about changing this:

> This means that even if you try to remove an event handler inside another one for the same event executed first, the removed handler will be executed anyway.

to this?

> Adding or removing event handlers on the current element won't take effect until the next time the event is handled. To prevent any further event handlers from executing on an element within an event handler, call `event.stopImmediatePropagation()`. 

Or something similar.

The thing that's strange about the W3C behavior is that it says `removeEventListener` takes effect immediately but `addEventListener` has no similar guarantee even though it could (it would have to be added on the end of the list). That asymmetry, plus the fact that nobody has complained in 9 years, tells me that this is pretty rare.

AurelioDeRosa:
I think your suggestions is fine. I'm going to update the PR so that we can merge it.

Issue #609, opened 2014-12-14
jQuery.ajax() supports "method" in preference to "type"


dmethvin:
Since version 1.9

http://bugs.jquery.com/ticket/12004
https://github.com/jquery/jquery/pull/1007

PR #671, opened 2015-02-28
Update jQuery.ajax.xml

AurelioDeRosa:
Fix issue #609


arthurvr:
I think we prefer `method` over `type` these days, so we should reflect that. All examples should be changed to use `method`, and `type` should be defined as a alias of `method`, not vice-versa.

AurelioDeRosa:
I'd say that it's `method` to be an alias as it was introduced later and for the reasons explained in the relative ticket. I agree that it's more descriptive of what the property means (HTTP method used) but I think there is also no plan for deprecate type? But this is my opinion, so I'd like to see what others have to say about it.

dmethvin:
For people switching between native and jQuery usage, the `method` makes more sense so I'd give preference to that. The term `type` has so many meanings that it doesn't explain much. However, I agree we haven't deprecated and wouldn't dare remove :smile_cat: the `type` property.

arthurvr:
Yeah, I think we should prefer `method` as well. @AurelioDeRosa Can you give this PR some updates to reflect that?

AurelioDeRosa:
PR updated.

arthurvr:
Wow. Looks good, @AurelioDeRosa. I'm wondering if we should update all of our examples too.

AurelioDeRosa:
I agree that we should update the examples too. I'll update the PR again.

arthurvr:
You rock, @AurelioDeRosa.

entries/jQuery.ajax.xml
```diff
@@ -104,6 +104,9 @@ $.ajax({
         <type name="Function"/>
         <desc>Specify the callback function name for a JSONP request.  This value will be used instead of the random name automatically generated by jQuery. It is preferable to let jQuery generate a unique name as it'll make it easier to manage the requests and provide callbacks and error handling. You may want to specify the callback when you want to enable better browser caching of GET requests. <strong>As of jQuery 1.5</strong>, you can also use a function for this setting, in which case the value of <code>jsonpCallback</code> is set to the return value of that function. </desc>
       </property>
+      <property default="'GET'" name="method" type="String" added="1.9.0">
+        <desc>The HTTP method to use for the request (e.g. "POST", "GET", "PUT"); default is "GET".</desc>
```
arthurvr: I know it was just takeover from the `type` option, but some nitpicks.

>  default is "GET".

Seems like weird to say this again, as we already indicate that just a line above that: http://gyazo.com/3e298875859e35d8dfdcef35f71ab3ac

> (e.g. "POST", "GET", "PUT")

Maybe wrap these in `<code>` elements.

arthurvr:
Looks good. Merging.

Issue #586, opened 2014-11-11
Docs for .data("key", undefined) are incorrect


rwaldron:
Originally reported by @vizmo here: https://github.com/jquery/jquery/issues/1852 

Re: http://api.jquery.com/data/

The incorrect bullet item: 

> `undefined` is not recognised as a data value. Calls such as `.data( "name", undefined )` will return the corresponding data for "name", and is therefore the same as `.data( "name" )`.

The correction: 

 `undefined` is not recognised as a data value. Calls such as `.data( "name", undefined )` will return the jQuery instance object that it was called on, which allows for method chaining. (Or something similar)

The correct behavior is supported by tests in the data test suite. https://github.com/jquery/jquery/blob/master/test/unit/data.js#L232

Issue #523, opened 2014-07-09
attr: Document attr(key, null) to remove attribute


Krinkle:
Our unit tests cover it explicitly, and the code implements it explicitly:
```js
		if ( value !== undefined ) {

			if ( value === null ) {
				jQuery.removeAttr( elem, name );
```

From the API perspective I'd also say this makes a lot of sense as it allows roundtripping between the getter returning null and the setter receiving null.

jugglinmike:
The discussion around the behavior of `attr` with `undefined`, while important, is unrelated to the intent of this issue. I think it should be safe to document "`null` unsets the attribute" even without consensus on `undefined`; is that right?

Krinkle:
@jugglinmike Yep. My main request here is to document (and thereby acknowledge) that `null` is supported.

Right now, passing `null` can appear fragile. Peers sometimes point it out when I this idiom in code – sometimes expressing a (very understandable) preference for messy ternaries that switch between `attr()` and `removeAttr()` based on null. I don't want my code to be responsible for this switch. It inevitably leads to inconsistency when people forget to do so. For separation of concerns, this should be (and already is) in jQuery. I just want to make sure I can rely on this (without risking a silently breaking change), and to have API documentation that I can reference as proof of sorts when discussing it with others.

Krinkle:
I still don't see this reflected on <https://api.jquery.com/attr/> ?

kswedberg:
@Krinkle, yeah... https://github.com/jquery/api.jquery.com/issues/890 
sigh.

PR #673, opened 2015-02-28
Update notes.xsl

AurelioDeRosa:
Fix issue #586.

### DIFF
--- a/notes.xsl
+++ b/notes.xsl
@@ -2,10 +2,10 @@
 <xsl:template name="note">
 	<xsl:choose>
 		<xsl:when test="@id = 'data-doesnt-accept-undefined'">
-			<code>undefined</code> is not recognised as a data value. Calls such as <code><xsl:value-of select="@data-title" />( <xsl:value-of select="@data-parameters" />, undefined )</code> will return the jQuery object that it was called on, allowing for chaining.
+			<code>undefined</code> is not recognized as a data value. Calls such as <code><xsl:value-of select="@data-title" />( <xsl:value-of select="@data-parameters" />, undefined )</code> will return the jQuery object that it was called on, allowing for chaining.
 		</xsl:when>
 		<xsl:when test="@id = 'jquery-data-doesnt-accept-undefined'">
-			<code>undefined</code> is not recognised as a data value. Calls such as <code><xsl:value-of select="@data-title" />( <xsl:value-of select="@data-parameters" />, undefined )</code> will return the corresponding data for "name", and is therefore the same as <code><xsl:value-of select="@data-title" />( <xsl:value-of select="@data-parameters" /> )</code>.
+			<code>undefined</code> is not recognized as a data value. Calls such as <code><xsl:value-of select="@data-title" />( <xsl:value-of select="@data-parameters" />, undefined )</code> will return the corresponding data for "name", and is therefore the same as <code><xsl:value-of select="@data-title" />( <xsl:value-of select="@data-parameters" /> )</code>.
 		</xsl:when>
 		<xsl:when test="@id = 'dimensions-number'">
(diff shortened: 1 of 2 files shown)


arthurvr:
The `data-doesnt-accept-undefined` note does [also appear](https://github.com/jquery/api.jquery.com/blob/bb5a21d6215d229190182b52bd77ffc98d046e23/entries/jQuery.data.xml#L28) on the `jQuery.data` post. Of course, `jQuery.data` [doesn't return the jQuery object](http://jsbin.com/goyajiwuhu/1/edit?js) is is called on.

AurelioDeRosa:
@arthurvr In this case I vote for splitting the two notes as they have a different behavior.

arthurvr:
Seems fine.

dmethvin:
Seems like we need to split the two, I agree. How do you guys feel about `undefined` not actually setting the value? 

> recognised

Have we standardised on UK English? Oh I mean, *standardized*. :smile: I guess I never noticed before.

AurelioDeRosa:
To be honest I feel more like `.data('thing', undefined)` should act as `.removeData('thing')`. We had a similar discussion in #523.

P.S. British English vs American English: their differences will kill me.

dmethvin:
I had forgotten about #523. It's tough because I think a good argument can be made either way. I agree with you that removing the data item feels best, but it would certainly be a breaking change as @gibson042 pointed out. It's hard to know whether a lot of plugins and developers depend on the current no-op behavior.

AurelioDeRosa:
@dmethvin What if we introduce this BC in jQuery 3.0?

dmethvin:
If very few people depend on this (or those people don't upgrade) then it would be nice to eliminate this kind of strange exception. If it turns out that several high-profile popular plugins break because we change this, then it would be a real pain to do and it would prevent people from upgrading to 3.0. I don't know which applies. :smile:

AurelioDeRosa:
@dmethvin we're all in your hands :)

dmethvin:
I'll leave the hard decisions to @timmywil now! Sucker!

arthurvr:
@AurelioDeRosa Besides from the discussion above, this PR needs some updates. Would you like to do them? I'd love to land this.

AurelioDeRosa:
@arthurvr I was waiting for a feedback by @timmywil.

timmywil:
Sorry I missed this. I think this can be landed. As far as #523, I think we'll need to gauge how much it will break. Personally, I always saw undefined as it's own value and thought it was most intuitive for `.data('something', undefined)` to act similarly to assigning a property on an object to the value `undefined`, and `.removeData('something')` being the equivalent of using the `delete` operator. Otherwise, determining whether something was set to `undefined` or whether it was never set is impossible. Regardless, we can thrash on that issue there.

arthurvr:
I can def. agree with @timmywil.

AurelioDeRosa:
So, I understand the conclusion is "Let's leave things as they are in regard of `undefined` and split the note in two notes". I'll update my PR accordingly.

arthurvr:
Thank you, @AurelioDeRosa!

arthurvr:
> I'll update my PR accordingly.

@AurelioDeRosa You've got the chance to update this PR? Would love to land it.

arthurvr:
ping @AurelioDeRosa

AurelioDeRosa:
PR updated @arthurvr

notes.xsl
```diff
@@ -2,6 +2,9 @@
 <xsl:template name="note">
 	<xsl:choose>
 		<xsl:when test="@id = 'data-doesnt-accept-undefined'">
+			<code>undefined</code> is not recognised as a data value. Calls such as <code><xsl:value-of select="@data-title" />( <xsl:value-of select="@data-parameters" />, undefined )</code> will return the jQuery object that it was called on, allowing for chaining.
```
arthurvr: > recognised

We use American English. Could you fix that on the while in the `jquery-data-doesnt-accept-undefined` note.

PR #674, opened 2015-03-01
css: add note about special meaning of mixed case

### LINKED ISSUE #357: Mixed case in .css() has specific meaning
So you can't use `.css("WiDtH")` for example.

http://bugs.jquery.com/ticket/14269

arthurvr:
Adds a quick note to the `css` entry about the special meaning of mixed case.

Would fix #357.


dmethvin:
LGTM!

PR #676, opened 2015-03-01
README: fix link for getting xmllint and xsltproc

### LINKED ISSUE #668: Replace GnuWin32 link
A while ago a user reported us an issue (#71) that caused the build to fail on Windows. @kswedberg fixed the documentation by updating the link with commit 6559302856cf3fecefa319d3d7ce1877c694b41b. The link was then changed again by @scottgonzalez with commit 04f42c47bfa8a3236b76dd69a332359384a1a4b5. However, I had the same issue of the user and ended up using the files at http://www.zlatkovic.com/libxml.en.html.

So, do you agree that we should replace the GnuWin32 link?

arthurvr:
Would fix #668. Once this one gets merged, I will submit a PR to every API repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ To build and deploy your changes for previewing in a [`jquery-wp-content`](https
 * [libxml2](http://xmlsoft.org/)
 * [libxslt](http://xmlsoft.org/libxslt/)
 
-The `xmllint` and `xsltproc` utilities need to be in your path. If you are on Windows, you can get libxml2 and libxslt from <a href="http://sourceforge.net/projects/gnuwin32/files/">GnuWin32</a>.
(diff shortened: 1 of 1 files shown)


arthurvr:
@scottgonzalez It would be nice if you could give this a +1 when it looks good. I'm not that familiar with our whole build system so would like another pair of eyes on it.

scottgonzalez:
I'm +1 based on other peoples' feedback. I don't use Windows, so I can't provide any actual feedback.

#669 would need to go into all repos as well.

arthurvr:
> I'm +1 based on other peoples' feedback

@AurelioDeRosa This worked for you right?

arthurvr:
> #669 would need to go into all repos as well.

I'll do that on the while.

AurelioDeRosa:
I had to download those files plus add the DLL as per #669 @arthurvr.

Issue #679, opened 2015-03-25
Selectors :hidden/:visible are slow and can force layout


dmethvin:
Need to add a note to the `:visible` and `:hidden` selector pages to the effect of:

> Using this selector heavily may cause performance issues because it can force the browser to re-render the page in order to determine visibility. Tracking the visibility of elements via other methods, for example a class name, can provide better performance.

There were some other optimizations [recently made in Sizzle](https://github.com/jquery/jquery/issues/2042) that improved performance but it still can be expensive.

dmethvin:
Ref https://github.com/jquery/jquery/issues/2108

arthurvr:
#682 :)

PR #682, opened 2015-03-26
hidden/visible-selector: add note about performance

arthurvr:
Would fix gh-679

### DIFF
--- a/notes.xsl
+++ b/notes.xsl
@@ -58,6 +58,9 @@
 		<xsl:when test="@id = 'html-code-execution'">
 			By design, any jQuery constructor or method that accepts an HTML string — <a href="/jQuery/">jQuery()</a>, <a href="/append/">.append()</a>, <a href="/after/">.after()</a>, etc. — can potentially execute code. This can occur by injection of script tags or use of HTML attributes that execute code (for example, <code>&lt;img onload=""&gt;</code>). Do not use these methods to insert strings obtained from untrusted sources such as URL query parameters, cookies, or form inputs. Doing so can introduce cross-site-scripting (XSS) vulnerabilities. Remove or escape any user input before adding content to the document.
 		</xsl:when>
+		<xsl:when test="@id = 'hidden-forces-layout'">
+			Using this selector heavily can have performance implications, as it may force the browser to re-render the page before it can determine visibility. Tracking the visibility of elements via other methods, using a class for example, can provide better performance.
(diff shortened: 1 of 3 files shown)


scottgonzalez:
In my opinion, this is too strongly worded. There are many sane uses of these selectors that don't warrant a strong warning.

dmethvin:
It's actually pretty weasel worded as it is. What alternate wording would you suggest?

notes.xsl
```diff
@@ -58,6 +58,9 @@
 		<xsl:when test="@id = 'html-code-execution'">
 			By design, any jQuery constructor or method that accepts an HTML string — <a href="/jQuery/">jQuery()</a>, <a href="/append/">.append()</a>, <a href="/after/">.after()</a>, etc. — can potentially execute code. This can occur by injection of script tags or use of HTML attributes that execute code (for example, <code>&lt;img onload=""&gt;</code>). Do not use these methods to insert strings obtained from untrusted sources such as URL query parameters, cookies, or form inputs. Doing so can introduce cross-site-scripting (XSS) vulnerabilities. Remove or escape any user input before adding content to the document.
 		</xsl:when>
+		<xsl:when test="@id = 'hidden-forces-layout'">
+			Using this selector heavily has some performance regressions, as it may force the browser to re-render the page before it can determine visibility. Tracking the visibility of elements via other methods, for example using a class, can provide better performance.
```
kswedberg: Not sure "regressions" is the right word here. What is it regressing from? Maybe, especially considering @scottgonzalez's concerns, we could change it to "implications"?

arthurvr: @kswedberg Yup. Will update it.

timmywil:
How about "can have performance implications" instead of "has some performance implications"?

scottgonzalez:
> How about "can have performance implications"

+1

arthurvr:
:+1:. I updated the PR. Looking good now?

scottgonzalez:
"Using this selector heavily" sounds odd. Do you mean things like using this inside a loop?

dmethvin:
It could be inside a loop where the DOM is being changed and thus forcing recalc each time, or could be just used once or twice in a very large DOM.

arthurvr:
Any suggestions are welcome, though I think it's quite okay.

dmethvin:
I think it's good now. :+1:

PR #684, opened 2015-03-27
jQuery.when: passing no arguments returns a resolved promise

### LINKED ISSUE #342: Document $.when() returns a resolved promise if given no arguments
When using $.when with an array (through `.apply`):

```js
var arrayOfDeferreds = [];
/* ...*/

$.when.apply( $, arrayOfDeferreds )
    .done( function () { } );
```

It works as expected even if the array remains empty (e.g. when aggregating return values of other parts of the application and collect promises) based on the way the code is currently implemented in jQuery core.

Could this behaviour be documented / officially supported (that giving it no arguments returns a resolved promise).

arthurvr:
Fixes #342 

@dmethvin @timmywil Any of you who wants to give this a quick review?

### DIFF
--- a/entries/jQuery.when.xml
+++ b/entries/jQuery.when.xml
@@ -21,6 +21,12 @@ $.when( { testing: 123 } ).done(function( x ) {
   alert( x.testing ); // Alerts "123"
 });
     </code></pre>
+    <p>If you don't pass it any arguments at all, <code>jQuery.when()</code> will return a resolved promise.</p>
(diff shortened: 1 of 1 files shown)


dmethvin:
Can you use `.then` there instead? Might as well start leaning towards the standard Promise methods.

arthurvr:
@dmethvin Of course, updated.

dmethvin:
:+1:

Issue #462, opened 2014-03-14
jQuery.parseHTML renders leading/trailing text nodes


dmethvin:
See http://bugs.jquery.com/ticket/14887 . This is correct behavior but worth mentioning in the docs.

arthurvr:
One of you who's willing to review #685?

PR #685, opened 2015-03-27
jQuery.parseHTML: add note about leading/trailing text nodes

arthurvr:
Fixes gh-462.

### DIFF
--- a/entries/jQuery.parseHTML.xml
+++ b/entries/jQuery.parseHTML.xml
@@ -15,7 +15,7 @@
     </argument>
   </signature>
   <longdesc>
-    <p><code>jQuery.parseHTML</code> uses a native DOM element creation function to convert the string to a set of DOM elements, which can then be inserted into the document.</p>
+    <p><code>jQuery.parseHTML</code> uses native methods to convert the string to a set of DOM nodes, which can then be inserted into the document. These methods do render all trailing or leading text (even if that's just whitespace). To prevent trailing/leading whitespace from being converted to text nodes you can pass the HTML string through <a href="/jQuery.trim/"><code>jQuery.trim</code></a>.</p>
     <p>By default, the <code>context</code> is the current <code>document</code> if not specified or given as <code>null</code> or <code>undefined</code>. If the HTML was to be used in another document such as an iframe, that frame's document could be used.</p>
 	<h2>Security Considerations</h2>
 	<p>Most jQuery APIs that accept HTML strings will run scripts that are included in the HTML. <code>jQuery.parseHTML</code> does not run script in the parsed HTML unless <code>keepScripts</code> is explicitly <code>true</code>. However, it is still possible in most environments to execute script indirectly, for example via the <code>&lt;img onerror&gt;</code> attribute. The caller should be aware of this and guard against it by cleaning or escaping any untrusted inputs from sources such as the URL or cookies. For future compatibility, callers should not depend on the ability to run <em>any</em> script content when <code>keepScripts</code> is unspecified or <code>false</code>.</p>


entries/jQuery.parseHTML.xml
```diff
@@ -17,6 +17,7 @@
   <longdesc>
     <p><code>jQuery.parseHTML</code> uses a native DOM element creation function to convert the string to a set of DOM elements, which can then be inserted into the document.</p>
     <p>By default, the <code>context</code> is the current <code>document</code> if not specified or given as <code>null</code> or <code>undefined</code>. If the HTML was to be used in another document such as an iframe, that frame's document could be used.</p>
+    <p>Worth noting is that <code>jQuery.parseHTML()</code>, unlike the standard jQuery selection methods, will actually render any leading or trailing text in the HTML string as text nodes (even if that's only whitespace). This is because jQuery uses a native method to transform the text into DOM elements. You can always prevent trailing/leading whitespace from getting rendered by passing the HTML string through <a href="/jQuery.trim/"><code>jQuery.trim</code></a>.</p>
```
dmethvin: I'd keep it a bit shorter by changing this:

> This is because jQuery uses a native method to transform the text into DOM elements. You can always prevent trailing/leading whitespace from getting rendered

to this:

> To prevent trailing/leading whitespace from being converted to text nodes,

The method we use isn't important enough to mention IMO.

arthurvr: Yeah, you're right, the majority of people doesn't care about that. Updated PR.

entries/jQuery.parseHTML.xml
```diff
@@ -17,6 +17,7 @@
   <longdesc>
     <p><code>jQuery.parseHTML</code> uses a native DOM element creation function to convert the string to a set of DOM elements, which can then be inserted into the document.</p>
```
gibson042: We can clarify things up front:
> … uses ~~a native DOM element creation function~~ **native functionality** to convert the string to a set of DOM ~~elements~~ **nodes** …

arthurvr: @gibson042 I noticed it but didn't change it as it was unrelated to this PR. I updated it now.

entries/jQuery.parseHTML.xml
```diff
@@ -17,6 +17,7 @@
...
     <p>By default, the <code>context</code> is the current <code>document</code> if not specified or given as <code>null</code> or <code>undefined</code>. If the HTML was to be used in another document such as an iframe, that frame's document could be used.</p>
+    <p>Worth noting is that <code>jQuery.parseHTML()</code>, unlike the standard jQuery selection methods, will actually render any leading or trailing text in the HTML string as text nodes (even if that's only whitespace). To prevent trailing/leading whitespace from being converted to text nodes you can pass the HTML string through <a href="/jQuery.trim/"><code>jQuery.trim</code></a>.</p>
```
gibson042: I'd drop the comparison to selection (which this is independent of) and move the meat of this paragraph to the end of the first (e.g., "to prevent trailing/leading whitespace from being appearing in text nodes, pass the HTML string through jQuery.trim").

arthurvr: I would agree with merging it into the second paragraph, though I like the explicit comparison. We know they're independent things but these are definitely things that people will confuse. I'll update my PR tomorrow.

entries/jQuery.parseHTML.xml
```diff
@@ -15,8 +15,9 @@
     </argument>
   </signature>
   <longdesc>
-    <p><code>jQuery.parseHTML</code> uses a native DOM element creation function to convert the string to a set of DOM elements, which can then be inserted into the document.</p>
+    <p><code>jQuery.parseHTML</code> uses a native functionality to convert the string to a set of DOM nodes, which can then be inserted into the document.</p>
```
kswedberg: @arthurvr, @dmethvin: What do you  think about changing "uses a native functionality" to "uses native DOM methods"?

arthurvr: Yeah, sounds better. Will do.

kswedberg: Thanks, @arthurvr !

arthurvr:
I played around with the wording a little an made the sentence a little more on-topic. Thanks for the suggestions guys. PTAL.

kswedberg:
LGTM

Issue #318, opened 2013-06-04
Wrong behavior described for multiple elements in append-ish methods


dmethvin:
> If there is more than one target element, however, cloned copies of the inserted element will be created for each target after the first.

should be

> If there is more than one target element, however, cloned copies of the inserted element will be created for each target except the last.

Affects `.append()`, `.prepend()`, `.appendTo()` etc.

arthurvr:
It looks like for `append()` and `prepend()` this is already fixed. I'm opening a PR for the other entries.

PR #687, opened 2015-03-28
appendTo/prependTo: correct note about appending multiple elements

arthurvr:
This already happened for `appendTo` and `prependTo`, so this fixes gh-318.


AurelioDeRosa:
I think an example of such behavior could be beneficial.

dmethvin:
:+1:

Issue #393, opened 2013-12-05
triggerHandler can also accept an Event, like trigger


hesselink:
In the documentation for `triggerHandler` it is stated that it 'behaves similarly to `trigger`'. But even though `triggerHandler` also accepts an `event` as the first argument, this is not a documented signature.

I think this signature should either be documented on `triggerHandler`, or it should be noted that it doesn't accept an `event` in the list of differences with `trigger` (in case this feature should not be used).

AurelioDeRosa:
The problem is that the documentation lacks the second signature. Anyone keen to confirm that it was added back in jQuery 1.3? If so, I'll take care of this issue.

dmethvin:
It's at least back to 1.3, see http://code.jquery.com/jquery-1.3.js , and it's not in 1.2.6.

arthurvr:
Finally this landed. Thanks for reporting this, @hesselink, sorry it took that long ;)

hesselink:
Thanks for the fix! Better late then never ;)

PR #690, opened 2015-03-31
triggerHandler: add signature for `.triggerHandler( event, extra )`

arthurvr:
Adds the `.triggerHandler( event, extraParameters )` signature to `.triggerHandler()`. Similar to the signature in the `.trigger()` entry.

Fixes https://github.com/jquery/api.jquery.com/issues/393.


dmethvin:
:+1:

Issue #686, opened 2015-03-28
jQuery.merge: add note about it correcting the `length` property


arthurvr:
> Internally we use jQuery.merge to merge jQuery objects which are array-like so they need to have length corrected. I see it's not documented at http://api.jquery.com/jQuery.merge/ so perhaps an API ticket is needed (cc @arthurvr)

https://github.com/jquery/jquery/issues/2155

mzgol:
This is documenting currently undocumented behavior so I'd like @jquery/core to approve it first.

markelog:
We say that `merge` excepts arrays not `array-like`s, therefore we would need to get funky of explaining why jquery corrects the length.

Do we need to document that? Essentially, we would be documenting `array-like`s as acceptable arguments, if user uses `merge` correctly, their wouldn't even know about that side-effect.

mzgol:
> Do we need to document that? Essentially, we would be documenting array-likes as acceptable arguments, if user uses merge correctly, their wouldn't even know about that side-effect.

The problem is `jQuery.merge` is public and de facto accepts array-likes; people use it to e.g. merge with the `arguments` object. So we might as well document this behavior, I doubt it'll ever change.

timmywil:
> we might as well document this behavior, I doubt it'll ever change.

Agreed. However, although we use this term in [several places already](http://api.jquery.com/?s=array-like), I think we should document what exactly "Array-like" means, perhaps [on the types page](http://api.jquery.com/Types/#Object).

mzgol:
@timmywil Good idea.

arthurvr:
Once the Array-like type lands to the types page I can update this PR. Thanks for the suggestions you all!

PR #692, opened 2015-03-31
jQuery.merge: add note about it correcting the `length`

arthurvr:
Fixes https://github.com/jquery/api.jquery.com/issues/686

### DIFF
--- a/entries/jQuery.merge.xml
+++ b/entries/jQuery.merge.xml
@@ -4,15 +4,15 @@
   <desc>Merge the contents of two arrays together into the first array. </desc>
   <signature>
     <added>1.0</added>
-    <argument name="first" type="Array">
-      <desc>The first array to merge, the elements of second added.</desc>
+    <argument name="first" type="ArrayLikeObjects">
+      <desc>The first array-like object to merge, the elements of second added.</desc>
     </argument>
-    <argument name="second" type="Array">
-      <desc>The second array to merge into the first, unaltered.</desc>
+    <argument name="second" type="ArrayLikeObjects">
+      <desc>The second array-like object to merge into the first, unaltered.</desc>
     </argument>
   </signature>
   <longdesc>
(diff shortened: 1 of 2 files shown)


arthurvr:
@mzgol WDYT?

mzgol:
The page still mentions that `jQuery.merge` merges two arrays (and does it in multiple places). This should all be changed to mention array-likes. Apart from that, the change looks fine (modulo the outcome of https://github.com/jquery/api.jquery.com/issues/686#issuecomment-90633413).

arthurvr:
So, I updated the entry to use the `array-like object` type so we don't need to get funky explaining why we correct the length. What do we think?

When updating it I noticed a little issue with https://github.com/jquery/api.jquery.com/commit/929f9e2472ec0da31446860f3e6a34f35321a6a2, I called the type `array-like objects` but it should be `array-like object`. Sorry! Fixed that in a separated commit.

mzgol:
I don't see it mentioned anywhere in the docs that if jQuery modifies an array-like object, its length gets corrected. This would fit perfectly here, though so I'd leave this text.

arthurvr:
@mzgol Updated that. Thanks! At first it sounded quite logical to me so I didn't find an explicit note necessary, but yeah, you're right. WDYT?

entries/jQuery.merge.xml
```diff
@@ -19,6 +19,7 @@ var newArray = $.merge([], oldArray);
     </code></pre>
     <p>This shortcut creates a new, empty array and merges the contents of oldArray into it, effectively cloning the array.</p>
     <p>Prior to jQuery 1.4, the arguments should be true Javascript Array objects; use <code>$.makeArray</code> if they are not.</p>
+    <p>Because <code>jQuery.merge()</code> is capable of merging array-like objects it will correct the <code>length</code> property.</p>
```
gibson042: Missing a comma before "it". But this note might be better placed near "is destructive" anyway, something like "It alters the <code>length</code> and numeric index properties of the first parameter to include items from the second."

arthurvr: @gibson042 Sounds good. PTAL.

gibson042:
LGTM.

PR #699, opened 2015-04-05
wrapAll: Updated return type of callback function

AurelioDeRosa:
The return value of the the callback function can be a jQuery object as well. This is the same as `wrap()`


arthurvr:
Looking good.

PR #703, opened 2015-04-06
innerWidth: return type unified

ericaro:
all (width,height, outerWidth, outerHeight, innHeight) dimensions have a return type "Number" but innerWidth which has a return type "Integer".

I don't see the reason for that, so I guess this is a small mistake.

Eric,

### DIFF
--- a/entries/innerWidth.xml
+++ b/entries/innerWidth.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <entries>
   <desc>Get the current computed inner width (including padding but not border) for the first element in the set of matched elements or set the inner width of every matched element.</desc>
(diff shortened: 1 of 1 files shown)


dmethvin:
:+1: , although it's not actually a mistake because the method originally could only return an integer. At this point there are cases where it may not be integer so we should clarify it.

arthurvr:
> At this point there are cases where it may not be integer so we should clarify it.

We should probably add a note about that, not just change it?

dmethvin:
It's not our change that made it more likely to return non-integer values, it's that screens are getting higher resolution and the browser is returning "fractional pixels" more often, for example.

Issue #708, opened 2015-04-08
Create an "Array-Like" type


dmethvin:
As suggested by @timmywil in https://github.com/jquery/api.jquery.com/issues/686#issuecomment-90646439, add an "Array-Like" type to our Types page to clarify when a method accepts more than just `Array`. Perhaps something like this:

> **Array-Like** - Either a true JavaScript `Array` or a JavaScript `Object` that contains a numeric `length` property. This latter case includes array-like objects commonly encountered in web-based code such as the `Arguments` object and the `NodeList` object returned by many DOM methods. Note that when a jQuery API accepts either plain Objects or Array-Like objects, a plain Object with a numeric `length` property will trigger the Array-Like behavior.

arthurvr:
Yup, :+1:. I'll take this one.

PR #710, opened 2015-04-09
Types: add new type `array-like objects`

arthurvr:
Fixes https://github.com/jquery/api.jquery.com/issues/708.

### DIFF
--- a/pages/Types.html
+++ b/pages/Types.html
@@ -73,6 +73,7 @@ var x = "";
       <li class="toclevel-2"><a href="#Array.3CType.3E_Notation"><span class="toctext">Array&lt;Type&gt; Notation</span></a></li>
     </ul>
   </li>
+  <li class="toclevel-1"><a href="#ArrayLikeObjects"><span class="toctext">Array-Like Objects</span></a></li>
   <li class="toclevel-1"><a href="#PlainObject"><span class="toctext">PlainObject</span></a></li>
   <li class="toclevel-1"><a href="#Date"><span class="toctext">Date</span></a></li>
   <li class="toclevel-1"><a href="#Function"><span class="toctext">Function</span></a>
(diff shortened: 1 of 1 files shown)


pages/Types.html
```diff
@@ -411,6 +412,9 @@ <h3 id="Array.3CType.3E_Notation"> Array&lt;Type&gt; Notation </h3>
 </pre>
 <p>This indicates that the method doesn't only expect an array as the argument, but also specifies the expected type. The notation is borrowed from Java 5's generics notation (or C++ templates).
 </p>
+<h2 id="ArrayLikeObjects">Array-Like Objects</h2>
+<p>Either a true JavaScript Array or a JavaScript Object that contains a numeric <code>length</code> property. This latter case includes array-like objects commonly encountered in web-based code such as the <code>arguments</code> object and the <code>NodeList</code> object returned by many DOM methods.</p>
```
markelog: We could a more strict here, not only `length` property should be in such object, but also their other properties names (but not necessary all properties) should be numeric too and they should corresponds with `length`.

Personally, i like Axel Rauschmayer definition - http://www.2ality.com/2013/05/quirk-array-like-objects.html

dmethvin: We don't enforce those extra rules though. A few days ago I read that ES6 defined "arraylike" as having a *non-negative* `length` value, which would be easy to enforce. However, I can't find any reference to that in the spec.

markelog: We can't iterate on non-numeric properties, whereas `merge` docs says

> Merge the contents of two arrays together into the first array.

if there would be `array-like` there, it should, according to this definition of `array-like` do this:
```js
var a = { test: 1 };
a.length = 1;

console.log( $.merge( [], a ).test ); // 1
```

since "test" property is also "content" of the array-like object.

gibson042: Our de facto definition, which I observe agrees very well with Axel Rauschmayer's, is a nonnegative numeric length and a property corresponding to the highest index: https://github.com/jquery/jquery/blob/2380028ec4a6a77401b867a51de26a3cb8e8d311/src/core.js#L434-L448

dmethvin: Oh, I remembered it backwards. Our definition checks for non-negative but I couldn't find that in the spec.

arthurvr: So, what's the consensus in here?

-

> Our de facto definition, which I observe agrees very well with Axel Rauschmayer's, is a nonnegative numeric length and a property corresponding to the highest index:

Using exactly that definition would suggest something like this works, isn't it?

```js
var obj = { test: 3, 1: 4 };
obj.length = 2;

console.log( isArraylike(obj) ); // true
console.log( $.merge( [], obj ).test ); // undefined
```

dmethvin: Yes I think by quoting @gibson042 you can't go wrong. :smile:

gibson042: I'd like this to be something like "an Object that contains a nonnegative integer <code>length</code> property and index properties from <code>0</code> up to <code>length - 1</code>". It describes the density that we assume (but are unwilling to fully verify at this point), and hopefully is clear enough to unambiguously include `{ length: 0 }`.

> Yes I think by quoting @gibson042 you can't go wrong.

I may hold you to that :wink:

markelog: And no one notice the guy who initially point that out :-)

arthurvr: That suggestion seems like reasonable @gibson042! Thanks y'all! I'll update the PR.

arthurvr:
Well, what do we think about this version?

gibson042:
:+1:

Also, :heart: @markelog for identifying both the problem and the solution. :smile:

dmethvin:
Just coming in to say that I :heart: @markelog too. Oh and the PR looks good too :smile_cat:

arthurvr:
:heart: :heartbeat: @markelog :heart_eyes_cat: :heart_decoration:

Issue #689, opened 2015-03-30
jQuery.fn.unwrap() now supports selector


timmywil:
See https://github.com/jquery/jquery/issues/1744

`.unwrap( selector )`

arthurvr:
Once I have a little of time I'll take this one. Thanks for opening the ticket Timmy.

arthurvr:
There we go: #711

arthurvr:
Fixed in v3 branch.

PR #711, opened 2015-04-09
unwrap: add `selector` argument

arthurvr:
PR for the v3 branch. Adds a `.unwrap(selector)` signature. Fixes https://github.com/jquery/api.jquery.com/issues/689.

Ref https://github.com/jquery/jquery/commit/7b09235ceed57bbcc26fc2c76147eb4e95ebdb92

### DIFF
--- a/entries/unwrap.xml
+++ b/entries/unwrap.xml
@@ -4,6 +4,12 @@
   <signature>
     <added>1.4</added>
   </signature>
+  <signature>
+    <added>3.0</added>
+    <argument name="selector" type="String" optional="true">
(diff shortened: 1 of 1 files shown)


entries/unwrap.xml
```diff
@@ -4,6 +4,12 @@
   <signature>
     <added>1.4</added>
   </signature>
+  <signature>
+    <added>3.0</added>
+    <argument name="selector" type="String" optional="true">
+      <desc>A selector to check the parent element against. If the parent element does not match the selector, the element won't be unwrapped.</desc>
```
dmethvin: There can be multiple elements so I'd change this to "If an element's parent does not" to emphasize that. Otherwise LGTM!

Issue #377, opened 2013-10-24
`.trigger()` and array-like arguments


timmywil:
Ported from http://bugs.jquery.com/ticket/14481

Something like...

"Note: When triggering with a plain object argument that is not "array-like" (as in a jQuery instance or an arguments object), but still contains a `length` property, pass the object in an array (e.g. `[ { length: 1 } ]`)"

arthurvr:
PR gh-716.

PR #716, opened 2015-04-16
Trigger: add note about objects with a `length` property

arthurvr:
Fixes gh-377. Once we land the new `array-like` type I can link that up from here as well.


arthurvr:
Anyone who wants to give this PR a quick review?

gibson042:
LGTM

Issue #725, opened 2015-04-28
Api for jQuery.post() shows "version added: 3.0"


ltongues:
![screenshot from 2015-04-28 20 21 27](https://cloud.githubusercontent.com/assets/7702679/7375890/4b40b7b8-ede4-11e4-9b45-be77cb0182cc.png)

dmethvin:
That information is actually correct, that particular signature won't be available until 3.0 ships. However it would be better to put the new signature after the old one, since that is the one most people will be seeking.

arthurvr:
> However it would be better to put the new signature after the old one, since that is the one most people will be seeking.

Yep, agreed, assigning to me :) Hope I'll have some time to do this this evening.

ltongues:
Ok, thanks.

PR #726, opened 2015-04-29
jQuery.post: change order of signatures

arthurvr:
Fixes gh-725

> However it would be better to put the new signature after the old one, since that is the one most people will be seeking.


dmethvin:
Yep, LGTM

Issue #709, opened 2015-04-09
jQuery(): add note about support of text nodes


arthurvr:
The default assumption should be that text nodes aren't supported by our APIs. We should add a note about it in the "using dom elements" section. Discussion started over at https://github.com/jquery/api.jquery.com/pull/592.

Perhaps we should also add it to `.contents()` which is another place where text nodes can sneak in.

arthurvr:
@pcj We're, as mentioned in the PR, still interested in hearing how the text node ended up in the collection.

arthurvr:
Would we be okay with explicitly saying something like

> Please note that you can pass text nodes into a jQuery collection this way, although, the default assumption should be that jQuery doesn't support manipulations on text nodes. The few APIs that support text nodes will have an explicit note about that on their API documentation     page.

on the `jQuery()` entry?

gibson042:
:+1: I like that a lot, but "manipulations" is overloaded in this context. "Operations" might be better, or explicitly mentioning the modules (is there anything other than traversing and filtering?).

arthurvr:
@gibson042 I totally agree. PR: https://github.com/jquery/api.jquery.com/pull/738. Mind taking a look and giving your LGTM if you like it?

timmywil:
Could we include comment nodes here? I think they mostly receive the same treatment as text nodes. See https://github.com/jquery/jquery/issues/2309.

Issue #149, opened 2012-10-16
Clarify that mixed types should not be passed to jQuery constructor


brandwaffle:
Pertains to: http://api.jquery.com/jQuery/

**User Reported:**
The page documents: 
jQuery( elementArray )
elementArray -- An array containing a set of DOM elements to wrap in a jQuery object.

However, as far as I can see, elementArray may also contain jQuery objects. This is a nice feature. It would be nice if it was documented! 

**Dev Summit Notes:**
Dave: incorrect assumption, however, it should be documented that passing jquery objects in the element array will cause them to be double-wrapped.  Such a child node would be singularly unloved - it would be an orphan forever, never to be adopted - it would live a life of dickensian tragedy.  Add docs saying NEVER PASS ANYTHING BUT DOM ELEMENTS via elementArray

PR #738, opened 2015-05-09
jQuery(): add note about support of text nodes

arthurvr:
Fixes https://github.com/jquery/api.jquery.com/issues/709

### DIFF
--- a/entries/contents.xml
+++ b/entries/contents.xml
@@ -6,7 +6,7 @@
   </signature>
   <desc>Get the children of each element in the set of matched elements, including text and comment nodes.</desc>
   <longdesc>
-    <p>Given a jQuery object that represents a set of DOM elements, the <code>.contents()</code> method allows us to search through the immediate children of these elements in the DOM tree and construct a new jQuery object from the matching elements. The <code>.contents()</code> and <code>.children()</code> methods are similar, except that the former includes text nodes as well as HTML elements in the resulting jQuery object.</p>
+    <p>Given a jQuery object that represents a set of DOM elements, the <code>.contents()</code> method allows us to search through the immediate children of these elements in the DOM tree and construct a new jQuery object from the matching elements. The <code>.contents()</code> and <code>.children()</code> methods are similar, except that the former includes text nodes, comment nodes as well as HTML elements in the resulting jQuery object. Please note that most jQuery operations don't support text nodes and comment nodes. The few that do will have an explicit note on their API documentation page.</p>
     <p>The <code>.contents()</code> method can also be used to get the content document of an iframe, if the iframe is on the same domain as the main page.</p>
     <p>Consider a simple <code>&lt;div&gt;</code> with a number of text nodes, each of which is separated by two line break elements (<code>&lt;br&gt;</code>):</p>
     <pre><code>
--- a/entries/jQuery.xml
+++ b/entries/jQuery.xml
@@ -57,7 +57,8 @@ $( "div.foo" ).click(function() {
       <p>Internally, selector context is implemented with the <code>.find()</code> method, so  <code>$( "span", this )</code> is equivalent to  <code>$( this ).find( "span" )</code>.</p>
 
       <h4 id="using-dom-elements">Using DOM elements</h4>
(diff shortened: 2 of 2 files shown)


entries/jQuery.xml
```diff
@@ -58,6 +58,7 @@ $( "div.foo" ).click(function() {
 
       <h4 id="using-dom-elements">Using DOM elements</h4>
       <p>The second and third formulations of this function create a jQuery object using one or more DOM elements that were already selected in some other way. When passing an array, each element <em>must</em> be a DOM element; mixed data is not supported. A jQuery object is created from the array elements in the order they appeared in the array; unlike most other multi-element jQuery operations, the elements are not sorted in DOM order.</p>
+      <p>Please note that you can pass text nodes into a jQuery collection this way, although, the default assumption should be that jQuery doesn't support operations on text nodes. The few APIs that support text nodes will have an explicit note about that on their API documentation page.</p>
```
gibson042: This contradicts some preceding text ("When passing an array, each element must be a DOM element; mixed data is not supported"), which IMO should just be removed because it's not true anyway. We might also be able to tighten up the new copy:
> Please note that although you can pass text nodes into a jQuery collection this way, most operations don't support them. The few that do will have an explicit note on their API documentation page.

entries/contents.xml
```diff
@@ -6,7 +6,7 @@
   </signature>
   <desc>Get the children of each element in the set of matched elements, including text and comment nodes.</desc>
   <longdesc>
-    <p>Given a jQuery object that represents a set of DOM elements, the <code>.contents()</code> method allows us to search through the immediate children of these elements in the DOM tree and construct a new jQuery object from the matching elements. The <code>.contents()</code> and <code>.children()</code> methods are similar, except that the former includes text nodes as well as HTML elements in the resulting jQuery object.</p>
+    <p>Given a jQuery object that represents a set of DOM elements, the <code>.contents()</code> method allows us to search through the immediate children of these elements in the DOM tree and construct a new jQuery object from the matching elements. The <code>.contents()</code> and <code>.children()</code> methods are similar, except that the former includes text nodes as well as HTML elements in the resulting jQuery object. Although this way text nodes can end up in a jQuery collection, please do keep in mind that only few operations work as expected on text nodes.</p>
```
gibson042: Copying the recommendation from below (i.e., I think the text should match):
> Please note that most operations don't support text nodes. The few that do will have an explicit note on their API documentation page.

gibson042:
Left a couple of suggestions, but I'm happy to see this and feel like it really resolves our stance on text nodes. :+1:

entries/jQuery.xml
kswedberg: @gibson042 : Take a look at https://github.com/jquery/api.jquery.com/issues/149. The part that begins with "Dave," I believe, is referring to @dmethvin.

gibson042: Ah, I remember that. We needed an admonition to not include jQuery collections _inside_ an array passed to the constructor. I'd still like this commit to remove the line about mixed data, and replace it with a note that elements of the array will be added to the collection as-is (and not unwrapped if they are jQuery collections).

arthurvr:
@timmywil As you asked I extended this explanation to include comment nodes.
@gibson042 Can you take another look and tell me what you think? I edited the things you asked for.

entries/contents.xml
```diff
@@ -6,7 +6,7 @@
   </signature>
   <desc>Get the children of each element in the set of matched elements, including text and comment nodes.</desc>
   <longdesc>
-    <p>Given a jQuery object that represents a set of DOM elements, the <code>.contents()</code> method allows us to search through the immediate children of these elements in the DOM tree and construct a new jQuery object from the matching elements. The <code>.contents()</code> and <code>.children()</code> methods are similar, except that the former includes text nodes as well as HTML elements in the resulting jQuery object.</p>
+    <p>Given a jQuery object that represents a set of DOM elements, the <code>.contents()</code> method allows us to search through the immediate children of these elements in the DOM tree and construct a new jQuery object from the matching elements. The <code>.contents()</code> and <code>.children()</code> methods are similar, except that the former includes text nodes, comment nodes as well as HTML elements in the resulting jQuery object. Please note that most jQuery operations don't support text nodes and comment nodes. The few that do will have an explicit note on their API documentation page.</p>
```
gibson042: > includes text nodes, comment nodes

I'd replace that comma with "and", but otherwise LGTM. Thanks!

arthurvr: Yep! I'll do that while merging.

PR #740, opened 2015-05-12
Clarified behavior of `:checked` selector. Turned text into link to `…

jhfrench:
…:selected` selector. Added example for how `:checked` selector works with `<option>` elements. Addresses https://github.com/jquery/api.jquery.com/issues/558

See [http://jsfiddle.net/jhfrench/n3w8y41p/](http://jsfiddle.net/jhfrench/n3w8y41p/) for a working demonstration of the new example.

### DIFF
--- a/entries/checked-selector.xml
+++ b/entries/checked-selector.xml
@@ -71,40 +71,6 @@ $( "input" ).on( "click", function() {
 </form>
 ]]></html>
   </example>
-  <example>
-    <desc>Determine how many input elements are checked.</desc>
-    <code><![CDATA[
-var countCheckedOptions = function() {
-  var n = $( "option:checked" ).length;
-  $( "#log_option" ).text( n + (n === 1 ? " is" : " are") + " ':checked'!" );
-};
-countCheckedOptions();
-
-$( "select" ).on( "change", countCheckedOptions );
-]]></code>
-    <css><![CDATA[
-  #log_option {
-    color: red;
-  }
-]]></css>
-    <html><![CDATA[
-<form>
-  <label for="cities">Cities I'd like to visit</label>
-  <br />
-  <select name="cities" id="cities" size="8" multiple="multiple">
-    <option value="Athens">Athens</option>
-    <option value="Barcelona">Barcelona</option>
-    <option value="Chicago">Chicago</option>
-    <option value="Detroit">Detroit</option>
-    <option value="Frankfurt">Frankfurt</option>
-    <option value="London">London</option>
-    <option value="Madrid">Madrid</option>
-    <option value="Paris">Paris</option>
-  </select>
-</form>
-<div id="log_option"></div>
-]]></html>
-  </example>
   <category slug="selectors/form-selectors"/>
   <category slug="version/1.0"/>
 </entry>
--- a/entries/checked-selector.xml
+++ b/entries/checked-selector.xml
@@ -7,7 +7,8 @@
   </signature>
   <desc>Matches all elements that are checked or selected.</desc>
   <longdesc>
-    <p>The <code>:checked</code> selector works for checkboxes, radio buttons, and select elements. For select elements only, use the <code>:selected</code> selector.</p>
+    <p>The <code>:checked</code> selector works for checkboxes, radio buttons, and "selected" options of select elements.</p>
+    <p>To retrieve only the selected options of <code>select</code> elements, use the <a href="/selected-selector/"><code>:selected</code></a> selector.</p>
   </longdesc>
   <example>
     <desc>Determine how many input elements are checked.</desc>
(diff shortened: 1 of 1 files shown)


entries/checked-selector.xml
```diff
@@ -7,7 +7,8 @@
   </signature>
   <desc>Matches all elements that are checked or selected.</desc>
   <longdesc>
-    <p>The <code>:checked</code> selector works for checkboxes, radio buttons, and select elements. For select elements only, use the <code>:selected</code> selector.</p>
+    <p>The <code>:checked</code> selector works for checkboxes, radio buttons, and "selected" options of select elements.</p>
```
gibson042: Why the quote marks around "selected"? Come to think of it, I think this mention of selection can be dropped completely.

arthurvr:
This is meant to fix https://github.com/jquery/api.jquery.com/issues/558.

arthurvr:
Thanks for your PR!

entries/checked-selector.xml
jhfrench: The original description is incorrect in that it claims `:selected` will return the `select` element that is "selected". The nuance is that the "selected" `option` element(s) is/are returned.

I'm using double quotes in an attempt to differentiate between the three versions of select:
`:select` --the selector
`select` --the element
`selected` --the `<option>` attribute

gibson042: > The original description is incorrect in that it claims :selected will return the select element that is "selected". The nuance is that the "selected" option element(s) is/are returned.

Right. I'm suggesting that the new second paragraph should explain `:selected` (the pseudo-selector), and the first paragraph doesn't mention selection at all:
> The <code>:checked</code> selector can match checkboxes, radio buttons, and options of select elements.

jhfrench: I'm certainly open for improvements, but I can't quite understand what you're suggesting. What do you propose for this sentence?

> The <code>:checked</code> selector works for checkboxes, radio buttons, and "selected" options of select elements.

gibson042: See above. I propose the following replacement:
> The <code>:checked</code> selector can match checkboxes, radio buttons, and options of `<select>` elements.

arthurvr: Hey-hey @jhfrench still interested in picking this up?

jhfrench: Yes--I will tr to do this today.

jhfrench:
@arthurvr I have signed the CLA as "Jeromy French jeromy_french@hotmail.com" also...what else do I need to do to get past this CLA check failure?

arthurvr:
@jhfrench The bot fails because you have two signatures on the sheet with the same email address. @scottgonzalez What's the best way to fix that?

jhfrench:
@arthurvr it's the "full name" label of that CLA sign-up form that threw me (and maybe others) off...made me think middle initial, which I rarely use, is expected.

scottgonzalez:
@arthurvr Right now the solution is to delete the duplicate, leaving the preferred signature in place. The long term solution is to allow multiple signatures, but as far as I know, nobody is working on anything related to the CLA checker.

arthurvr:
> Right now the solution is to delete the duplicate, leaving the preferred signature in place. 

@scottgonzalez Who can do that for us?

entries/checked-selector.xml
```diff
@@ -70,6 +71,40 @@ $( "input" ).on( "click", function() {
...
+    <option value="Detroit">Detroit</option>
+    <option value="Frankfurt">Frankfurt</option>
+    <option value="London">London</option>
+    <option value="Madrid">Madrid</option>
+    <option value="Paris">Paris</option>
+  </select>
+</form>
+<div id="log_option"></div>
+]]></html>
+  </example>
```
arthurvr: This example is unrelated to this commit (and to the issue you're trying to fix. Also I personally don't believe it adds that much value.

jhfrench: @arthurvr How do you you mean? The example demonstrates how the `:checked` selector can be used with `option` elements; this demonstration is missing from the current documentation.

Do you want me to separate into two tickets: #558 and a new ticket for giving an example of using `:checked` with an `option` element?

arthurvr: Well, I understand what the example is trying to demonstrate I just don't really think it adds enough value to be there. To me this looks like a really clear entry that doesn't really need a big example. What does the rest think?

AurelioDeRosa: To be honest I agree with @arthurvr. I don't think there is a need for an example here.

dmethvin: I think select-multiple is a pretty rare thing anyway so an example based on it doesn't add a lot of practical value. There is always a tradeoff between adding content to clarify usage and making the page so long that nobody takes the time to read it. I like to err on the side of brevity until there is some common misconception that needs to be corrected.

jhfrench: I'll make the case for the additional example like this: of the three HTML elements with which the `:checked` selector can be used, the are examples for two of them. This additional example covers the remaining, missing, third use-case (`option` elements).

That said, I will bow to the wisdom of the team. :)

arthurvr: Nah. Not worth it IMHO, I think the page is pretty clear as-is. If everybody agrees, can you remove the example from the commit? Thanks.

arthurvr:
Looks like the CLA bot is alright with this now. Thanks.

jhfrench:
So...we good here? What's left to be done?

AurelioDeRosa:
Everything is fine. I'm going to merge it tonight.

PR #743, opened 2015-05-23
jQuery.xml: Wrapped code using the `code` element

AurelioDeRosa:


arthurvr:
Thanks. Commit components don't need an extension, by the way :)

AurelioDeRosa:
I'll keep that in mind @arthurvr. Thank you

Issue #712, opened 2015-04-09
Clarify description of .val( array )


gibson042:
Ref jquery/jquery#2195

The paragraph on https://api.jquery.com/val/#val-value starting with "Passing an array of element values" is inaccurate, and less clear than it could be. In actuality, we implement two similar but distinct special behaviors for array input, both effectively updating the affected collection to the intersection of provided input with possible values:
* `checkbox` and `radio`: check/uncheck based on whether or not their value is in the array
* `select` elements: select/unselect options based on whether or not their value is in the array

Although jQuery does not directly touch elements outside the context collection, the `.val` setter _can_ affect them by triggering native behavior restricting to 1 the maximum count of checked radio inputs in a group or selected options in a select-one.

PR #744, opened 2015-05-23
val(): Updated description for the setter version

AurelioDeRosa:
Fixes: gh-712


gibson042:
> the <code>option</code> whose value appears last in the array will be selected.

I'm not sure we want to lock ourselves into that behavior. It's not terribly likely to change, but it's also not a great idea to pass in junk and let us sort it out. I'd rather not dignify (or encourage) it with recognition.

AurelioDeRosa:
Thank you for the feedback @gibson042. Anyone wants to comment before I update the PR?

AurelioDeRosa:
PR updated.

AurelioDeRosa:
Anyone keen to review this PR?

gnarf:
One question I have about this behavior after reading the changed text:

* Will `$( "input.checkboxes" ).val( anArrayOfValues )` set `elem.checked = false` if the value is not in the array as well?

AurelioDeRosa:
@gnarf Yes, it does. I'll update the PR accordingly.

AurelioDeRosa:
Can you double check this PR @arthurvr?

PR #753, opened 2015-06-08
Promise objects also have the .promise method.

ChadKillingsworth:
I realize that code implementation does not necessarily dictate documentation, but the Promise type returned from various methods also contains the `.promise()` method of the deferred object (and I think it has always been this way).

This would simply officially document that behavior.

Discussion which prompted this: https://groups.google.com/d/msg/closure-compiler-discuss/qbNLjzp3OCw/fjNhLsv4regJ


dmethvin:
I agree, this LGTM. In general we are moving towards interoperation with standard Promise and thenables, but there is no reason I can think of to avoid documenting `.promise()` for completeness.

arthurvr:
Thanks! :+1: I'll land this one.

Issue #730, opened 2015-05-04
Document .data changes


gibson042:
jquery/jquery/issues/2257

As of 3.0, data methods camelCase all keys before access. Use of non-camelCase keys requires direct object manipulation (e.g., `var obj = $el.data()`).

AurelioDeRosa:
Closed with commit 270b51818cf5e07f70b1d0cd1689f83518579663.

PR #758, opened 2015-06-14
data: Document behavior changes

AurelioDeRosa:
Closes: gh-730


entries/data.xml
```diff
@@ -31,6 +31,7 @@ $( "body" ).data(); // { foo: 52, bar: { myType: "test", count: 40 }, baz: [ 1,
       </code></pre>
       <p>In jQuery 1.4.3 setting an element's data object with <code>.data(obj)</code> extends the data previously stored with that element.</p>
       <p>Prior to jQuery 1.4.3 (starting in jQuery 1.4) the <code>.data()</code> method completely replaced all data, instead of just extending the data object. If you are using third-party plugins it may not be advisable to completely replace the element's data object, since plugins may have also set data.</p>
+      <p>jQuery 3 changes the behavior of this method to align it to the <a href="http://www.w3.org/TR/html5/dom.html#dom-dataset">Dataset API specifications</a>. Specifically, jQuery 3 transforms all the properties' key to be camelCase. Writing a statement like <code>$( "body" ).data( {"my-name": "aValue"} ).data();</code> will return <code>{ myName: "aValue" }</code>.</p>
```
arthurvr: Follow the jQuery style guide: `{"my-name": "aValue"}` => `{ "my-name": "aValue" }`

arthurvr: ditto in the other sentence.

mzgol: Perhaps it'd be useful to mention that not everything is camelCased, only `/-([a-z])/ -> $1`, e.g. `a-b-3-c` is changed to `aB-3C`.

entries/data.xml
```diff
@@ -88,6 +89,7 @@ alert( $( "body" ).data( "foo" ) ); // undefined
 $( "body" ).data( "bar", "foobar" );
 alert( $( "body" ).data( "bar" ) ); // foobar
       </code></pre>
+      <p>jQuery 3 changes the behavior of this method to align it to the <a href="http://www.w3.org/TR/html5/dom.html#dom-dataset">Dataset API specifications</a>. Specifically, jQuery 3 transforms all the properties' key to be camelCase. Writing a statement like <code>$( "body" ).data( {"my-name": "aValue"} ).data();</code> will return <code>{ myName: "aValue" }</code>.</p>
```
arthurvr: Mostly we wrap important version differences in `<strong>` tags (or at least the version).

> Dataset API specifications

Not sure myself, should this be "specification" or "specifications"

entries/data.xml
AurelioDeRosa: @mzgol Do you have a suggestion on the wording here? I think many users won't really like or understand the use of a regex in the description.

arthurvr:
Why does CI fail?

arthurvr:
oh nevermind. We don't have a travis config in the v3 branch yet.

AurelioDeRosa:
@arthurvr I was wondering what was the reason as well.

arthurvr:
Simply because our Travis CI config isn't in the V3 branch yet. I'll rebase that branch this evening.

entries/data.xml
gibson042: Every two-character sequence of "-" (U+002D) followed by a lowercase ASCII letter is replaced by the uppercase version of the letter. It's the algorithm defined at [http://www.w3.org/TR/html5/dom.html#dom-dataset](http://www.w3.org/TR/html5/dom.html#dom-dataset).

mzgol: Yep, if we want to be specific here (and IMO we should, even if we just add a note explaining what camelCasing means here), what @gibson042 said is the most accurate.

AurelioDeRosa:
Anyone keen to review so that we can merge this PR? // cc @arthurvr

entries/data.xml
```diff
@@ -88,6 +89,7 @@ alert( $( "body" ).data( "foo" ) ); // undefined
 $( "body" ).data( "bar", "foobar" );
 alert( $( "body" ).data( "bar" ) ); // foobar
       </code></pre>
+      <p><strong>jQuery 3</strong> changes the behavior of this method to align it to the <a href="http://www.w3.org/TR/html5/dom.html#dom-dataset">Dataset API specifications</a>. Specifically, jQuery 3 transforms all the properties' key to be camelCase. Writing a statement like <code>$( "body" ).data( { "my-name": "aValue" } ).data();</code> will return <code>{ myName: "aValue" }</code>.</p>
```
mzgol: Shouldn't this note be changed in the same way as the upper one?

AurelioDeRosa: Oh...yes. I'll do it now.

mzgol: Is there a way to avoid duplication, define it in one place & just use in these two? We do have includes but preferably the pattern would be defined here since it's not needed on other pages. Is it possible?

AurelioDeRosa: Apart from the global notes I'm not aware of any other method, sorry.

AurelioDeRosa:
Closed with commit 270b51818cf5e07f70b1d0cd1689f83518579663.

PR #779, opened 2015-07-20
deferred.always: Added note for parameters changing order

### LINKED ISSUE #763: deferred.always parameters changing based on success?
Hello,

I am reading this on the jquery.ajax documentation page:

```
jqXHR.always(function( data|jqXHR, textStatus, jqXHR|errorThrown ) { });
An alternative construct to the complete callback option, the .always() method replaces the deprecated .complete() method.
```

This is a documentation error or a bad joke, right?
The parameter order changing based on wether the query succeeds or not?

How is one to determine the actual status code without actually knowing if it was successful or not?

AurelioDeRosa:
Fixes #763


arthurvr:
@AurelioDeRosa Same thing here. There's no point with doing PRs if you don't wait for a reviewer and an LGTM...

PR #782, opened 2015-07-23
Build: move redirects from the infrastucture repo

arthurvr:
Possible as of grunt-jquery-content 2.3.0, and now we want to port all redirects over from the nginx configs to the repos. cc @gnarf 

Ref https://github.com/jquery/infrastructure/blob/puppet-master/modules/jquery/manifests/wp/jquery.pp#L66. Apparently for this site there's only one redirect that needs to be moved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,6 @@
   },
   "dependencies": {
     "grunt": "0.4.5",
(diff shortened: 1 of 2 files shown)


gnarf:
:+1:

Issue #737, opened 2015-05-08
New :visible and :hidden definitions


gibson042:
jquery/jquery/issues/2227

As of 3.0, elements will be considered `:visible` if they have any layout boxes, including those of zero width and/or height (e.g., `<br>` elements and inline elements with no content).

AurelioDeRosa:
Fixed in branch v3 with commit d0e7282d4c0e9c02263ebcb15ecd419a25113941.

PR #789, opened 2015-07-27
Visible: Specified new behavior

AurelioDeRosa:
Fixes gh-737


arthurvr:
This also affects `:hidden` so let's also add a note on that entry.

AurelioDeRosa:
You didn't give me the time :)

PR updated.

AurelioDeRosa:
Merged with commit d0e7282d4c0e9c02263ebcb15ecd419a25113941.

mzgol:
Why "will be"? Isn't this supposed to go live when jQuery 3 is released when the future tense won't make sense?

arthurvr:
@AurelioDeRosa We all open PRs and wait for code reviews before landing something and I'd prefer if you did so too. That's the point of PRs :)

AurelioDeRosa:
You saw this PR after ~30 seconds it was made and the only comment you had was

> This also affects :hidden so let's also add a note on that entry.

I took it as a LGTM. Simple as that, there is no need to paternalize. :)

arthurvr:
> you saw this PR after ~30 seconds it was made and the only comment you had was

Yes, because I'm not going to review a PR when it's not complete yet. Just like everybody does, please wait for an LGTM, :+1: or alike comment.

arthurvr:
And for https://github.com/jquery/api.jquery.com/pull/779, you didn't have any comment yet ;)

Issue #479, opened 2014-04-22
Global Ajax events work only with document


AurelioDeRosa:
Hello.
All the global Ajax event handlers have the following note:

> As of jQuery 1.8, the .ajaxStart() method should only be attached to document.

I want to focus on the word "should" which is not valid anymore. While this was true for jQuery 1.8, in newer versions of jQuery you can only listen for these events on `document`. Therefore the documentation should be updated with an integrative note (at least this is my opinion).

This issue is related with #118.

AurelioDeRosa:
To fix this issue I propose to create a global note that is added to all Ajax global event. Any thoughts?

dmethvin:
Agreed.

arthurvr:
Sounds good to me :+1:

PR #797, opened 2015-08-16
Added note for Global Ajax Event Handlers

AurelioDeRosa:
Fixes gh-479

### DIFF
--- a/notes.xsl
+++ b/notes.xsl
@@ -67,6 +67,9 @@
 		<xsl:when test="@id = 'detach-shorthand'">
 			As the <code>.<xsl:value-of select="@data-event"/>()</code> method is just a shorthand for <code>.on( "<xsl:value-of select="@data-event"/>", handler )</code>, detaching is possible using <code>.off( "<xsl:value-of select="@data-event"/>" )</code>.
 		</xsl:when>
+		<xsl:when test="@id = 'global-ajax-event'">
+			Starting with jQuery 1.9, all the handlers for the <a href="/category/ajax/global-ajax-event-handlers/">jQuery global Ajax events</a>, including those added with the <code><xsl:value-of select="@data-title"/></code> method, must be attached to <code>document</code>.
+		</xsl:when>
 	</xsl:choose>
 </xsl:template>
 </xsl:stylesheet>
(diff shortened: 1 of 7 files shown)


notes.xsl
```diff
@@ -67,6 +67,9 @@
 		<xsl:when test="@id = 'detach-shorthand'">
 			As the <code>.<xsl:value-of select="@data-event"/>()</code> method is just a shorthand for <code>.on( "<xsl:value-of select="@data-event"/>", handler )</code>, detaching is possible using <code>.off( "<xsl:value-of select="@data-event"/>" )</code>.
 		</xsl:when>
+		<xsl:when test="@id = 'global-ajax-event'">
+			Starting with jQuery 1.9, all the handlers for the <a href="/category/ajax/global-ajax-event-handlers/">jQuery global Ajax events</a>, including those added with the <code><xsl:value-of select="@data-title"/></code> method, must be attached to <code>document</code>.
```
kswedberg: Is it 1.9 or 1.8? Descriptions currently say 1.8.

kswedberg:
Thanks @AurelioDeRosa. Are you going to also remove the notes that currently appear in the `<longdesc>`? (for example, "As of jQuery 1.8, the <code>.ajaxStop()</code> method should only be attached to <code>document</code>.")

notes.xsl
AurelioDeRosa: In 1.8 you should, in 1.9 you must. Is that clear enough?

AurelioDeRosa:
I think your comment is related to https://github.com/jquery/api.jquery.com/issues/798. Can you check that issue I raised?

kswedberg:
Yeah, that's what I saw. What I'm suggesting is, if you're adding the "global" note to each method, you should also remove the "local" note.

notes.xsl
kswedberg: Ah, I see. I think what you have is fine then. We typically use "As of jQuery x.x" instead of "Starting with x.x." Would you mind changing that phrase?

AurelioDeRosa: Updated based on your suggestion :)

AurelioDeRosa:
In this case, should I highlight that from 1.8 is recommended, but from 1.9 is mandatory?

kswedberg:
Never mind. I (finally) see what you're getting at. Maybe just change the global note from "must" to "_must_" (using `<em></em>`).

If you do that, then it LGTM, so feel free to merge it..

AurelioDeRosa:
@kswedberg updated based on your comment :)

kswedberg:
Excellent. Thanks, @AurelioDeRosa ! Wanna merge it in?

AurelioDeRosa:
@kswedberg Done!

kswedberg:
Thanks, @AurelioDeRosa !

PR #800, opened 2015-08-26
the type of argument can be an ArrayLikeObject

yadimon:
not only `array` can be accepted, but `ArrayLikeObject` too:

```javascript
var $selected = $('.selected');
$.grep($selected, function(e, i){return true;})
```


yadimon:
edited through github directly -> wrong CLA :/

PR #801, opened 2015-08-28
the type of argument can be an ArrayLikeObject

yadimon:
`grep`:
not only `array` can be accepted, but `ArrayLikeObject` too:

```javascript
var $selected = $('.selected');
$.grep($selected, function(e, i){return true;})
```

if you look at the implementation
https://github.com/jquery/jquery/blob/2.1-stable/src/core.js#L384
it should be fully supported.
like by merge:
https://github.com/jquery/jquery/blob/2.1-stable/src/core.js#L370

### DIFF
--- a/entries/jQuery.grep.xml
+++ b/entries/jQuery.grep.xml
@@ -4,7 +4,7 @@
   <desc>Finds the elements of an array which satisfy a filter function. The original array is not affected.</desc>
   <signature>
     <added>1.0</added>
-    <argument name="array" type="Array">
+    <argument name="array" type="ArrayLikeObject">
       <desc>The array to search through.</desc>
     </argument>
     <argument name="function" type="Function">
--- a/entries/jQuery.grep.xml
+++ b/entries/jQuery.grep.xml
@@ -5,7 +5,7 @@
   <signature>
     <added>1.0</added>
     <argument name="array" type="ArrayLikeObject">
-      <desc>The array to search through.</desc>
+      <desc>The array-like object to search through.</desc>
     </argument>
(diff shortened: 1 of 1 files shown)


arthurvr:
@timmywil Is this officially supported or simply a handy side effect?

gnarf:
`jQuery.grep` is [only tested against real arrays](https://github.com/jquery/jquery/blob/aabe94edb4880c75eeebc5b5b5d66a9ad17008fe/test/unit/core.js#L946-L964)

However it is passed a jQuery object (`ArrayLikeObject`) inside the code for [`jQuery.fn.filter`](https://github.com/jquery/jquery/blob/a644101ed04d0beacea864ce805e0c4f86ba1cd1/src/traversing/findFilter.js#L78)

I'd say this is expected and :+1: on landing this update.  Though we should also add a test case for filtering some array like? cc @timmywil

AurelioDeRosa:
Once https://github.com/jquery/jquery/pull/2605 is merged, I'll merge this PR.

entries/jQuery.grep.xml
```diff
@@ -4,7 +4,7 @@
   <desc>Finds the elements of an array which satisfy a filter function. The original array is not affected.</desc>
   <signature>
     <added>1.0</added>
-    <argument name="array" type="Array">
+    <argument name="array" type="ArrayLikeObject">
       <desc>The array to search through.</desc>
```
AurelioDeRosa: @yadimon Can you improve this sentence by specifying that it can be also an Array-like object?

PR #802, opened 2015-09-07
Completed the list of copied props of jQuery.Event.

andyli:
Source ref: https://github.com/jquery/jquery/blob/master/src/event.js (props, keyHooks.props, mouseHooks.props)

### DIFF
--- a/categories.xml
+++ b/categories.xml
@@ -143,7 +143,7 @@ jQuery( "body" ).trigger( e );
             </li>
           </ul>
           <p>The following properties are also copied to the event object, though some of their values may be undefined depending on the event:</p>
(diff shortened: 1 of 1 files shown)


dmethvin:
This LGTM. It's adding the standard mouse and key events to the "also copied" list but since we say they can be `undefined` that covers the case for things like mouse properties on a key event well enough.  Thanks @andyli!

AurelioDeRosa:
It looks good to me as well, thank you @andyli. Should I merge it?

kswedberg:
@AurelioDeRosa , yes, please merge away! :+1:

arthurvr:
:+1: Thanks!

PR #804, opened 2015-09-09
Effects: Clarify that callbacks are per-element

### LINKED ISSUE #803: Clarify `always` callback for animations
The docs are unclear that `always` is also a per-element callback, like `complete`.

https://github.com/jquery/jquery/issues/2585#issuecomment-138949538

dmethvin:
Fixes #803


AurelioDeRosa:
The PR looks good to me and can be merged.

I agree that we should create a global note that is included in all the files affected. I suggest to merge this PR and then open a new issue that takes into account the general problem.

PR #807, opened 2015-09-21
Added page about the contextmenu() alias

### LINKED ISSUE #806: Documentation about contextmenu() is missing
The website doesn't have a documentation page for the `contextmenu()` alias which is supported since the early days of jQuery.

AurelioDeRosa:
Fixes #806

### DIFF
--- a/entries/contextmenu.xml
+++ b/entries/contextmenu.xml
@@ -1,0 +1,86 @@
+<?xml version="1.0"?>
+<entry type="method" name="contextmenu" return="jQuery">
+  <title>.contextmenu()</title>
+  <desc>Bind an event handler to the "contextmenu" JavaScript event, or trigger that event on an element.</desc>
+  <signature>
+    <added>1.0</added>
+    <argument name="handler" type="Function">
+      <desc>A function to execute each time the event is triggered.</desc>
+      <argument name="eventObject" type="Event" />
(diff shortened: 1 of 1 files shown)


AurelioDeRosa:
Anyone keen to review again the PR?

arthurvr:
Would appreciate if someone from the core team could take a look too just to make sure we don't miss out on anything. cc @timmywil @dmethvin @markelog @gibson042

markelog:
It is common perception in Core team, that event aliases are tricky shouldn't be used so they might be even deprecated in the future, also it looks like we don't have any tests for it  - https://github.com/jquery/jquery/issues/2636

arthurvr:
Yeah, I know. I always prefer to avoid them too but still it's better to have them at least documented. Maybe we should add some kind of note that tells people to avoid these methods though.

markelog:
> Maybe we should add some kind of note that tells people to avoid these methods though.

That sounds like deprecation notice :-)

AurelioDeRosa:
Based on your comments, I'd say let's add the note *if* and *when* these shortcuts will be deprecated.

AurelioDeRosa:
@kswedberg can you perform a final check please?

kswedberg:
LGTM! :+1:

PR #811, opened 2015-09-23
Add code example to jQuery.ajax `accepts` property

dbazile:
The current version of this documentation section only says that this property accepts an object and makes no mention of the structure of that object.

All credit to [Darin Dimitrov's answer to this StackOverflow question](http://stackoverflow.com/a/11065286).


AurelioDeRosa:
Hi @dbazile.

Thank you for your contribution. Can you perform a couple of changes so that I can merge your PR?

Update this text:

> in the <em>Accept</em> request header that tells the server what kind of response it will accept in return.

to:

> in the <code>Accept</code> request header. This header tells the server what kind of response it will accept in return.

Also, after this sentence you have two spaces instead of one (before "For example").

Finally, can you change:

> for this to function properly.

to:

> for this to work properly.

dbazile:
Done and done!

AurelioDeRosa:
LGTM. Thank you

Issue #312, opened 2013-05-27
addClass/removeClass change the property not the attribute


dmethvin:
Clarify that these two methods manipulate the property and not the attribute. In HTML DOM docs the browser echoes the `className` property value back into the `class` attribute. Two implications:

* The browser may not remove the `class` attribute when the last class name is removed, it may only be set to an empty string.
* The `addClass/removeClass` methods only work for documents with HTML DOM semantics (e.g., _not_ pure XML documents).

PR #813, opened 2015-10-06
Document that addClass and removeClass change the property

AurelioDeRosa:
Fixes gh-312

### DIFF
--- a/entries/addClass.xml
+++ b/entries/addClass.xml
@@ -19,7 +19,7 @@
   <desc>Adds the specified class(es) to each element in the set of matched elements.</desc>
   <longdesc>
     <p>It's important to note that this method does not replace a class. It simply adds the class, appending it to any which may already be assigned to the elements.</p>
-    <p><code>.addClass()</code> manipulates the property of the selected elements, not the attribute. Once the property is changed, it's the browser that updates the <code>class</code> attribute accordingly. An implication of this behaviour is that this method only works for documents with HTML DOM semantics (e.g., not pure XML documents).</p>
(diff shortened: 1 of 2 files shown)


AurelioDeRosa:
@agcolom @kswedberg Fixed! Thank you for the feedback.

entries/addClass.xml
```diff
@@ -19,6 +19,7 @@
   <desc>Adds the specified class(es) to each element in the set of matched elements.</desc>
   <longdesc>
     <p>It's important to note that this method does not replace a class. It simply adds the class, appending it to any which may already be assigned to the elements.</p>
+    <p><code>.addClass()</code> manipulates the property of the selected elements, not the attribute. Once the property is changed, it's the browser that updates the <code>class</code> attribute accordingly. An implication of this behavior is that this method only works for documents with HTML DOM semantics (e.g., not pure XML documents).</p>
```
arthurvr: I wonder if it makes sense to also mention which property we're actually talking about, instead of just "the property"

kswedberg: @arthurvr, @AurelioDeRosa: :+1: to specifying the property. Maybe something like this?…

The <code>.addClass()</code> method manipulates the `className` *property* of the selected elements, not the `class` *attribute*. Once the property is changed, it's the browser that updates the attribute accordingly.

gibson042: Note also that behavior will change in jQuery 3.0, where we will set the `class` attribute instead of the `className` property for better SVG support: https://github.com/jquery/jquery/issues/2199

AurelioDeRosa:
PR updated. @gibson042 I'll open a separate issue for that and create a different PR if you agree. This way we can merge it into the relevant branch.

gibson042:
:+1:

AurelioDeRosa:
Anyone keen to a final review for this?

dmethvin:
LGTM, and we will revisit it when 3.0 lands.

PR #834, opened 2015-11-18
noConflict: Update return type to reflect actual behavior

mmiszy:


scottgonzalez:
This change is actually incorrect. The [`jQuery` type](http://api.jquery.com/Types/#jQuery) represents an `jQuery` *instance*. The return value from `jQuery.noConclift()` is the `jQuery` constructor, not an instance.

johnnyreilly:
Hi @scottgonzalez,

Thanks for replying to @mmiszy. Is there a way to improve this PR?  Is there an approved way to refer to the `jQuery` function in the API docs?  It seems a shame to refer to it as just `Object` when we know it's something distinct from that.

scottgonzalez:
This is not. I believe this is the only method that would have such a return value, besides the obsolete `jQuery.sub()` method, which actually has the wrong return value listed.

@timmywil Are you interested in adding a new type to avoid one-off changes for things like TypeScript type definitions?

AurelioDeRosa:
Friendly ping @timmywil

Issue #835, opened 2015-11-19
noConflict - second example need call moved


drummike:
In the second code snippet shown on the noConflict page there are these lines:
<snip>
$.noConflict();
jQuery( document ).ready(function( $ ) {
  // Code that uses jQuery's $ can follow here.
});
// Code that uses other library's $ can follow here.
</snip>
The noConflict call should be moved down, so it happens after the line with ready(function($) 
Otherwise $ is undefined, or defined as belonging to some other library.

I made a fork with the change to entries/jQuery.noConflict.xml

mzgol:
No, this example is correct. `$.noConflict()` restores the previous global value of `$` but not `jQuery`. The .ready function gets a parameter that's assigned to the jQuery instance so `$` inside of the ready callback is equal to the first jQuery. This is a function parameter and not the global `$`.

drummike:
Are you certain? The reason I ask is that I tried that code and it gave me an error.

<!DOCTYPE html>
<html lang="en">
<body>
<script src="../js/jquery.js"></script>
<script>
$.noConflict();
jQuery( document ).ready(function( $ ) {
  // Code that uses jQuery's $ can follow here.
  $.alert("that dog don't hunt");  // line 9 - for reference
});
// Code that uses other library's $ can follow here.

</script>
</body>
</html>
== console output
Uncaught TypeError: $.alert is not a function

scottgonzalez:
Well, `$.alert` is not a function that *should* exist given that code, since there is no such method in jQuery.

drummike:
Duh! Oh well, a PBCAK.

Cheers.

PR #836, opened 2015-11-19
Moved noConflict call to below where $ is used. One line change.

drummike:
As described in the issue I opened on this subject and in the git tag, it was a one line change. noConflict was called in a code snippet just before where $ was used. The noConflict line was moved down below the point where $ was used.


mzgol:
See https://github.com/jquery/api.jquery.com/issues/835#issuecomment-158003496.

PR #839, opened 2015-11-25
Typo on line 80

BrachaTenenbaum:
"makes" changed to "make"


mzgol:
Thanks but the original text is OK and your PR would introduce a mistake.

PR #842, opened 2015-12-03
Add description to fig

agcolom:


entries/animate.xml
```diff
@@ -58,7 +58,7 @@ $( "#clickme" ).click(function() {
     <p class="image">
       <figure>
         <img src="/resources/animate-1.jpg" alt=""/>
-        <figcaption>figure 1</figcaption>
+        <figcaption>Figure 1 - Illustration of the specified animation effect</figcaption>
```
AurelioDeRosa: Instead of "specified" can be use "described"? It sounds a bit better to me.

entries/fadeIn.xml
```diff
@@ -40,7 +40,7 @@ $( "#clickme" ).click(function() {
       <img class="column three" src="/resources/0042_06_34.png" alt=""/>
       <img class="column three" src="/resources/0042_06_35.png" alt=""/>
       <img class="column three" src="/resources/0042_06_36.png" alt=""/>
-      <figcaption>figure 1</figcaption>
+      <figcaption>Figure 1 - Illustration of the <code>fadeIn()</code> effect</figcaption>
```
AurelioDeRosa: On @kswedberg commit I commented:

> I'm not an expert but I wonder if from the accessibility point of view is confusing to have a `figure` with one `figcaption` and multiple images.

Worth checking.

entries/animate.xml
agcolom: I used specified as in "specified in the code"...

entries/slideToggle.xml
```diff
@@ -42,7 +42,7 @@ $( "#clickme" ).click(function() {
       <img class="column three" src="/resources/0042_06_26.png" alt=""/>
       <img class="column three" src="/resources/0042_06_27.png" alt=""/>
       <img class="column three" src="/resources/0042_06_28.png" alt=""/>
-      <figcaption>figure 1</figcaption>
+      <figcaption>Figure 1 - Illustration of the <code>slideToggle()</code> effect</figcaption>
```
AurelioDeRosa: "...effect when hiding the image"?

entries/slideToggle.xml
```diff
@@ -50,7 +50,7 @@ $( "#clickme" ).click(function() {
       <img class="column three" src="/resources/0042_06_30.png" alt=""/>
       <img class="column three" src="/resources/0042_06_31.png" alt=""/>
       <img class="column three" src="/resources/0042_06_32.png" alt=""/>
-      <figcaption>figure 2</figcaption>
+      <figcaption>Figure 2 - Illustration of the <code>slideToggle()</code> effect</figcaption>
```
AurelioDeRosa: "...effect when showing the image"?

entries/toggle.xml
```diff
@@ -64,7 +64,7 @@ $( "#clickme" ).click(function() {
       <img class="column three" src="/resources/0042_06_10.png" alt=""/>
       <img class="column three" src="/resources/0042_06_11.png" alt=""/>
       <img class="column three" src="/resources/0042_06_12.png" alt=""/>
-      <figcaption>figure 1</figcaption>
+      <figcaption>Figure 1 - Illustration of the <code>toggle()</code> effect</figcaption>
```
AurelioDeRosa: Same as `fadeToggle()`

entries/toggle.xml
```diff
@@ -72,7 +72,7 @@ $( "#clickme" ).click(function() {
       <img class="column three" src="/resources/0042_06_14.png" alt=""/>
       <img class="column three" src="/resources/0042_06_15.png" alt=""/>
       <img class="column three" src="/resources/0042_06_16.png" alt=""/>
-      <figcaption>figure 2</figcaption>
+      <figcaption>Figure 2 - Illustration of the <code>toggle()</code> effect</figcaption>
```
AurelioDeRosa: Same as `fadeToggle()`

entries/fadeIn.xml
agcolom: I think this is ok, since the group of images is composing the one figure, only the figure needs a figcaption. http://html5doctor.com/the-figure-figcaption-elements/ mentions this case for example.

entries/slideToggle.xml
kswedberg: :+1:

entries/slideToggle.xml
kswedberg: :+1:

entries/fadeIn.xml
kswedberg: +1 to @agcolom . There's an example of multiple images at http://www.w3.org/WAI/tutorials/images/groups/ too. It has nested figures/figcaptions, so the same might not apply in our case, but it suggests using `role="group"` and an `aria-labelledby` attribute. Your call.

entries/slideToggle.xml
agcolom: @kswedberg the :+1: goes to which comment? (my entry or Aurelio's suggestion?)
Thanks :-)

kswedberg: sorry, @agcolom :) It goes with @AurelioDeRosa's suggestion.

entries/fadeIn.xml
AurelioDeRosa: Thank you for the heads up. I learnt something new.

AurelioDeRosa:
@agcolom I know you're busy, so let me know if I can help with this PR. Do you want me to complete it?

entries/fadeIn.xml
AurelioDeRosa: After an additional check, I think the case doesn't apply to this example. So, I'll leave it as per @agcolom commit.

AurelioDeRosa:
Completed the PR and merged as discussed in our last meeting, @agcolom.

PR #845, opened 2015-12-12
index() returns an integer

andyli:
Just to be accurate ;)
The integer type is described in https://api.jquery.com/Types/#Integer.
Several methods/properties already use integer as return type. e.g. [length](https://api.jquery.com/length/)


dmethvin:
Good point, since this is an index it can't be fractional. LGTM.

PR #855, opened 2015-12-29
Indicate where data will be passed in $.one doc

cmisenas:
I only found the `data` after inspecting `event` further. This change will make it clearer where to get `data` (and consistent with the docs for `.one( events [, selector ] [, data ], handler )` and `.one( events [, selector ] [, data ] )`).

If people think it would be great to make this change for other jquery fns too, the following fns still use the old desc:

```
entries/bind.xml
entries/blur.xml
entries/change.xml
entries/click.xml
entries/contextmenu.xml
entries/dblclick.xml
entries/delegate.xml
entries/error.xml
entries/focus.xml
entries/focusin.xml
entries/focusout.xml
entries/keydown.xml
entries/keypress.xml
entries/keyup.xml
entries/live.xml
entries/load-event.xml
entries/mousedown.xml
entries/mouseenter.xml
entries/mouseleave.xml
entries/mousemove.xml
entries/mouseout.xml
entries/mouseover.xml
entries/mouseup.xml
entries/resize.xml
entries/scroll.xml
entries/select.xml
entries/submit.xml
```

I'd be happy to make changes as part of this PR

### DIFF
--- a/entries/one.xml
+++ b/entries/one.xml
@@ -8,7 +8,7 @@
       <desc>A string containing one or more JavaScript event types, such as "click" or "submit," or custom event names.</desc>
     </argument>
     <argument name="data" type="PlainObject" optional="true">
-      <desc>An object containing data that will be passed to the event handler.</desc>
+      <desc>Data to be passed to the handler in <a href="/event.data/"><code>event.data</code></a> when an event is triggered.</desc>
     </argument>
     <argument name="handler" type="Function">
       <desc>A function to execute at the time the event is triggered.</desc>


dmethvin:
The link definitely helps clarify things!

As for all the now-discouraged "shortcut" methods, I would still prefer they just be defined in terms of the non-deprecated `.on` method rather than duplicating all the details about event handlers across more than a dozen pages. 

Right now the text for the shortcut methods is all copy/pasta/edit, but I think it could be done with a `notes.xml` entry added to all the shortcut pages to the effect of:

> This method is simply a shortcut for the use of the `.on()` method. Specifically, `.METHOD(handler)` is the same as `.on( "METHOD", handler )`; `.METHOD( data, handler )` is the same as `.on( "METHOD", data, handler )`; and `.METHOD()` is the same as `.trigger( "METHOD" )`. For more information about events, see the entry for the `.on()` method.

Where METHOD is the title of that particular page, like `<xsl:value-of select="@data-title" />`. By doing it this way we can make the edit in that page once and tweak it without touching so many pages in the future.

AurelioDeRosa:
Hi @cmisenas and thank you for your contribution. The PR looks good to me.

Some notes:

- @cmisenas would you like to submit another PR to address @dmethvin point?
- @dmethvin I think it's still worth, at least for now, to update the documentation in all of those pages. Doing so, we keep the documentation more consistent and accurate.

dmethvin:
Sure @AurelioDeRosa, whatever works. It definitely affects them all but for whatever reason it seems like the selector in `.prev` and `.next` confuses people the most.

mgol:
Landed, thanks @cmisenas!

PR #859, opened 2016-01-07
Manipulation: fix after/before xml parsing

timmywil:
I ran `grunt deploy` locally and ran into this error. Not sure why it's not happening in CI testing. This update seems to fix it.

### DIFF
--- a/entries/after.xml
+++ b/entries/after.xml
@@ -30,9 +30,9 @@
       </return>
     </argument>
   </signature>
-   <signature>
+  <signature>
     <added>1.10</added>
-    <argument name="function" type="Function">
+    <argument name="function-html" type="Function">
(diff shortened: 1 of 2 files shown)


AurelioDeRosa:
What operating system are you using? If it's Windows, I can try to replicate your issue. Also, what's the error?

timmywil:
I'm using Mac 10.11.2. Error is...

```
Running "build-xml-entries:all" (build-xml-entries) task
>> Error: element a: validity error : ID after-function already defined
```

AurelioDeRosa:
PR looks good to me, so I can merge. However, before doing it if anyone with a Mac could try it, we can understand the reason behind the error.

timmywil:
I don't think there's a danger in merging. I'm surprised Windows users don't see this error. There are duplicate names.

Issue #851, opened 2015-12-16
Document that .toggleClass( boolean ) has been deprecated in jQuery 3


AurelioDeRosa:
Reference: https://github.com/jquery/jquery/pull/2618.

dmethvin:
Given that is wasn't really documented before (just a signature with no real explanation) I wonder if we should relegate this to the release notes.

timmywil:
We've decided to save this removal for 4.0. We just need to mark it deprecated in 3.0.

dmethvin:
Oh I missed that I guess. Yes, it's still in master so deprecate in 3.0 and remove in 4.0 then.

AurelioDeRosa:
No worries Dave.

PR #860, opened 2016-01-08
toggleClass: Document deprecation of a signature

AurelioDeRosa:
Fixes gh-851

### DIFF
--- a/entries/toggleClass.xml
+++ b/entries/toggleClass.xml
@@ -1,67 +1,62 @@
 <?xml version="1.0"?>
-<entry type="method" name="toggleClass" return="jQuery">
-  <title>.toggleClass()</title>
-  <signature>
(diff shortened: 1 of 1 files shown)


dmethvin:
Not sure how we want to document this, but the signature is actually removed as of 3.0. Perhaps we document it as deprecated as of 1.12.0?

timmywil:
That's one way to go, but I figured we'd remove it in 4.0 https://github.com/jquery/api.jquery.com/issues/851#issuecomment-169804944.

AurelioDeRosa:
Is it good to merge?

timmywil:
It looks ok to me, but the build is failing.

AurelioDeRosa:
We have the usual issue that branch 3 doesn't have the Travis file, thus it fails.

Issue #857, opened 2016-01-04
Clarify and correct .closest() docs


dmethvin:
https://github.com/jquery/jquery/issues/2796

Remove reference to `this.context` ("context of the jQuery set") since it has been deprecated and was removed in 3.0.

Clarify that the `context` argument is used for jQuery-extension positional selectors such as `:first` and `nth()` but otherwise only for deciding how far up the tree to stop comparing nodes . That is, the context is not used as the root for matching the selector.

gibson042:
> Clarify that the `context` argument is used for jQuery-extension positional selectors such as `:first` and `nth()` but otherwise only for deciding how far up the tree to stop comparing nodes . That is, the context is not used as the root for matching the selector.

I'd like to take a different tack and [ensure that positionals never match](https://github.com/jquery/jquery/pull/2818), in which case this issue is satisfied by removing the "If no context is passed in…" sentence.

PR #874, opened 2016-01-18
jQuery.closest: Removed reference to deprecated context property

AurelioDeRosa:
Fixes gh-857


AurelioDeRosa:
Note: I'm targeting master instead of the branch v3 because the property was already deprecated in version 2.

agcolom:
LGTM

PR #877, opened 2016-01-23
Added .editorconfig file

AurelioDeRosa:
`.editorconfig` file taken directly from jQuery's repository.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,27 @@
+# This file is for unifying the coding style for different editors and IDEs
+# editorconfig.org
+
+root = true
+
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+# Tabs in JS unless otherwise specified
+[**.js]
+indent_style = tab
+
+[test/**.xml]
+indent_style = tab
+
+[test/**.php]
+indent_style = tab
+
+[test/**.html]
+indent_style = tab
+
+[test/**.css]
+indent_style = tab
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,14 +14,8 @@ insert_final_newline = true
 [**.js]
 indent_style = tab
 
-[test/**.xml]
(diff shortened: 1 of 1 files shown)


scottgonzalez:
It's kinda silly to do a direct copy from a different project with different types of files and a different directory structure. Half of these rules are for files that don't even exist, and the majority of the files in this repo don't even get indentation rules applied to them.

AurelioDeRosa:
For the benefit of solving this issue, it'd be great if you could point on specific lines that you'd like to change/add/remove. By doing so, we can have a change that works for everyone and  improve the _status quo_. In its current state your comment doesn't really help.

kswedberg:
@AurelioDeRosa: why bother specifying `[**/.js]`, `[**.xml]`, etc.? Is there a type of file in which we don't use `indent_style = tab`? I think having one `indent_style = tab` along with the other rules should be sufficient. Or am I missing something?

scottgonzalez:
> I think having one indent_style = tab along with the other rules should be sufficient. Or am I missing something?

Right, almost everything is tab indented, so having that be the default would make sense. `package.json` and `*.yml` should be set to two spaces though.

.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+root = true
+
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+# Tabs in JS unless otherwise specified
```
scottgonzalez: There is nothing else specified.

mgol:
ping @AurelioDeRosa

mgol:
@AurelioDeRosa It seems to me that [current version of Core's `.editorconfig`](https://github.com/jquery/jquery/blob/43a0b5e/.editorconfig) would be a good fit, the only modification needed is to apply 2 spaces to `*.yml` files as well as @scottgonzalez indicated.

Would you like to do it? If not, I'd close this PR and I can do it separately.

Issue #876, opened 2016-01-20
Better explanation for when keydown event is triggered


acshef:
The documentation at https://api.jquery.com/keydown/ states,
> The `keydown` event is sent to an element when the user first presses a key on the keyboard.

This is not the full story. The keydown event also fires as long as the key is being held (hold a key in the example at the bottom of that documentation page to see the handler repeatedly get called).

Perhaps there needs to be distinction between "pressing a key on the keyboard" (hardware-level) and the *browser* actually getting a keyboard event at the OS/software-level.

AurelioDeRosa:
Hi @acshef. I agree that the sentence could be clearer and it's worth a clarification. Do you want to submit a pull request for it?

PR #892, opened 2016-02-16
Clarified when the keydown event is fired

AurelioDeRosa:
Fixes gh-876


agcolom:
LGTM

AurelioDeRosa:
Thank you for the quick review @agcolom.

PR #900, opened 2016-03-12
jQuery.ajax: Improved description of the jsonp property

AurelioDeRosa:


dmethvin:
Oh good catch that we didn't document the `false` case. The "security reasons" may not be clear to everyone but hey, we warned them. In general I'm not sure that many devs understand how JSONP works. :smile_cat:  But this LGTM.

AurelioDeRosa:
Thank you for the review @dmethvin. FWIW, I agree with you about the understanding of JSONP.

Issue #910, opened 2016-04-13
Attribute Equals Selector docs may be vague


ericandrewlewis:
The [Attribute Equals Selector](https://api.jquery.com/attribute-equals-selector/) doc says that the attribute value "Can be either an unquoted single word or a quoted string."

This is a bit vague in light of [jquery/jquery#2824](https://github.com/jquery/jquery/issues/2824#issuecomment-171483952), as `#anchor` needs to be quoted. Perhaps we should just say that the value must be quoted, as describing how `#anchor` is a special case and not a single word seems a bit complex.

georgestephanis:
The w3 attribute selector specification requires that attribute values must be CSS **identifiers **or **strings**. -- https://www.w3.org/TR/css3-selectors/#attribute-selectors

In this context, **identifiers **may be comprised exclusively of case-insensitive alphanumeric characters, hyphens, underscores, and ISO 10646 characters U+00A0 and higher.  As the # sign does not fall under this subset of characters (it's actually U+0023), the attribute value need to use the **string **syntax -- and in this context, strings must always be quoted.

dmethvin:
I think an "unquoted single word" is pretty clear, so `abc` and `dog` are words but `#abc` and `~$&@*` are not. If there is a simpler small phrase we can use there that doesn't require a paragraph let us know. Remember that this definition would need to be repeated anywhere that the identifier/string distinction is present in CSS selectors, including all the attribute selectors and not just this one. The spec's "CSS identifier" isn't that clear unless you follow the link and read the paragraph that describes it. 

The basic problem here is that jQuery first tried the high-performance `querySelectorAll` method, which rightly threw an error. When it used Sizzle to parse the string it got through via a bug and the incorrect selector was allowed, doing what the author intended 99 percent of the time.

So this wording is really all about justifying our right to make an invalid and undocumented case break the way it should have all along. It shouldn't be a problem with new code because people would most likely figure it out quickly.

aaronjorbin:
Unquoted single word on its own isn't necessarily true.  While `#abc` isn't a word, `Ke$ha` is a word (even though she decided to switch to Kesha).

Perhaps "Can be either an unquoted single word (only containing alphanumeric characters) or a quoted string." ?  Makes it clearer what word means.

georgestephanis:
But the spec permits hyphens and underscores as well 🕵

dmethvin:
[...]

PR #911, opened 2016-04-14
Link to W3C for unquoted single word

aaronjorbin:
Unquoted single word has a specific definition in this case that is not
succinct. A link to the spec helps developers understand what is meant.

fixes gh-910
ref jquery/jquery#2824


dmethvin:
This LGTM, thanks for catching all the attribute selectors!

AurelioDeRosa:
LGTM as well. Merging.

Issue #908, opened 2016-04-12
category not found


feiwen8772:
http://api.jquery.com/category/version/1.12/2.2/ this page not found。

/version/1.12&2.2/ maybe suitable。

AurelioDeRosa:
This issue can be found at [$.uniqueSort()](https://api.jquery.com/jQuery.uniqueSort/). It was introduced in [commit 82c1f4c](https://github.com/jquery/api.jquery.com/commit/82c1f4c57d31142a4851d50cad4a0233a8c27036). @timmywil do you have any suggestion on the action here?

timmywil:
1.12&2.2 works for me.

AurelioDeRosa:
Is our system able to deal with the `&` character? If not, we could use a dash.

dmethvin:
Dash sounds safest.

timmywil:
Dash makes it look like 1.12 to 2.2, tho.

dmethvin:
True. If `&` or `&amp;` works I'm down with it but I always get a bad feeling about character entities.

kswedberg:
@timmywil , @AurelioDeRosa : When I add the `1.12&amp;2.2` slug to categories.xml, I get `Warning: Invalid slug: 1.12&2.2.`

AurelioDeRosa:
So, I had a bit of time to dig into this issue. The `&` doesn't work and the dash is the only character I managed to have working. Is everyone happy with it?

dmethvin:
Dash works for me. You could say my hopes were dashed. Or not.

kswedberg:
@AurelioDeRosa Works for me!

mgol:
@kswedberg You need to add a new tag (`npm version patch` & push) to have the changes reflected on the site.

kswedberg:
@mgol right you are! Sorry, dropped the ball there.

kswedberg:
Hmm. seems you already did it. Thanks, @mgol !

mgol:
Yup, I just wanted to remind you about it for the sake of the future. :)

-- 
Michał Gołębiowski

PR #913, opened 2016-04-19
jQuery.uniqueSort: Fixed category name

AurelioDeRosa:
Fixes gh-908

### DIFF
--- a/categories.xml
+++ b/categories.xml
@@ -414,6 +414,14 @@ var files = event.originalEvent.dataTransfer.files;
         <hr/>
       ]]></desc>
     </category>
+    <category name="Version 1.12 &amp; 2.2" slug="1.12-2.2">
+      <desc><![CDATA[
+        <p>Aspects of the API that were changed in the corresponding versions of jQuery. Changes in jQuery 1.12 and 2.2 includes performance improvements of the selector engine, manipulation of class names for SVG elements, support for the Symbol type and iterators added in ES2015, and a new hook has been added for filtering HTML. A <a href="https://github.com/jquery/jquery-migrate">jQuery Migrate Plugin</a> was offered to help developers with a transitional upgrade path.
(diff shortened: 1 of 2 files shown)


dmethvin:
Ah, so the encoded version works? Awesome!

AurelioDeRosa:
@agcolom did you have the chance to double-check this?

agcolom:
@AurelioDeRosa My local setup is no longer working for the core api (not sure what change broke it yet, and I cannot install anything as I get the error on npm install!) so I haven't been able to test this, sorry.

kswedberg:
@AurelioDeRosa this needs a corresponding `<category>` entry in the categories.xml file. Unfortunately, when I try to add `<category name="Version 1.12&amp;2.2" slug="1.12&amp;2.2">…</category>`, I get "Warning: Invalid slug: 1.12&2.2." Using a hyphen (-) works. Maybe we should use that instead?

kswedberg:
@AurelioDeRosa ugh, I think I spoke too soon. While I don't get the error when I use `<category name="Version 1.12-2.2" slug="1.12-2.2">` (hyphens). The build process seems to ignore it. Maybe there's a stricter check when adding the version than when validating it?

AurelioDeRosa:
@kswedberg I managed to have the dash working. I think what was wrong is that `$.uniqueSort()` doesn't not include the category at the bottom of the file.

If everyone is happy, I can update the PR and merge it.

kswedberg:
Aha! You're correct, @AurelioDeRosa . That was the missing piece of the puzzle. Thanks so much for looking into this. Please update and merge.

AurelioDeRosa:
PR updated. I double checked and everything works correctly. As soon as I get approval, I'll merge it.

mgol:
LGTM

Issue #904, opened 2016-03-28
Ajax example on when docs will fail with git build / 3.0


arschmitz:
https://api.jquery.com/jquery.when/

The first example resolves with multiple arguments which will fail on the git build and in upcoming 3.0 release because it uses then which no longer allows multiple args and will resolve with an array now in this case. 

It's also a bad example as it lack error handling

AurelioDeRosa:
I think we should keep the example because people might rely on it and maintainers of projects using our previous versions might want to understand the pattern. My proposal is to add a `fail()` call to the existing example and add a change in branch v3 with a single argument. We should  also add an explanation of what is changed.

dmethvin:
The only reason I'd want to keep it is to explain that people should not do it. That's fine though. Any call like `$.when( something-that-is-always-a-Deferred )` doesn't need `$.when` in the first place so we can explain that.

PR #915, opened 2016-04-19
jQuery.when: Improved example

AurelioDeRosa:
Fixes gh-904

### DIFF
--- a/entries/jQuery.when.xml
+++ b/entries/jQuery.when.xml
@@ -12,10 +12,15 @@
     <p>If no arguments are passed to <code>jQuery.when()</code>, it will return a resolved Promise.</p>
     <p>If a single Deferred is passed to <code>jQuery.when()</code>, its Promise object (a subset of the Deferred methods) is returned by the method. Additional methods of the Promise object can be called to attach callbacks, such as <a href="/deferred.then/"><code>deferred.then</code></a>. When the Deferred is resolved or rejected, usually by the code that created the Deferred originally, the appropriate callbacks will be called. For example, the jqXHR object returned by <code>jQuery.ajax()</code> is a Promise-compatible object and can be used this way:</p>
     <pre><code>
-$.when( $.ajax( "test.aspx" ) ).then(function( data, textStatus, jqXHR ) {
-  alert( jqXHR.status ); // Alerts 200
+$.when( $.ajax( "test.aspx" ), $.ajax( "page.aspx" ) ).then(function( res1, res2 ) {
+  alert( res1[2].status ); // Alerts 200
+})
+.fail(function( err1, err2 ) {
+  alert( err1.status );
 });
     </code></pre>
+    <p>In the previous example, <code>res1</code> and <code>res2</code> are arrays containing three values: <code>data</code>, <code>textStatus</code>, and <code>jqXHR</code>. It is adviced to always provide a callback to handle errors through the use of <code>deferred.fail()</code> or the second parameter of <code>deferred.then()</code>.</p>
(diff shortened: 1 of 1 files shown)


entries/jQuery.when.xml
```diff
@@ -12,10 +12,15 @@
     <p>If no arguments are passed to <code>jQuery.when()</code>, it will return a resolved Promise.</p>
     <p>If a single Deferred is passed to <code>jQuery.when()</code>, its Promise object (a subset of the Deferred methods) is returned by the method. Additional methods of the Promise object can be called to attach callbacks, such as <a href="/deferred.then/"><code>deferred.then</code></a>. When the Deferred is resolved or rejected, usually by the code that created the Deferred originally, the appropriate callbacks will be called. For example, the jqXHR object returned by <code>jQuery.ajax()</code> is a Promise-compatible object and can be used this way:</p>
     <pre><code>
-$.when( $.ajax( "test.aspx" ) ).then(function( data, textStatus, jqXHR ) {
-  alert( jqXHR.status ); // Alerts 200
+$.when( $.ajax( "test.aspx" ), $.ajax( "page.aspx" ) ).then(function( res1, res2 ) {
+  alert( res1[2].status ); // Alerts 200
+})
+.fail(function( err1, err2 ) {
```
dmethvin: Should we switch to `.catch()` here @gibson042? Also since the failure occurs as soon as any rejection occurs, I think it's possible that `res1` does not yet have a status if `res2` rejected. If a programming exception occurred they might even be `Error` objects instead of `jQXHR` objects! In real code you'd need to inspect both to sort it out.

Perhaps for the purposes of this example we could just `alert("one or more requests failed")`.

gibson042: Your instincts are good, @dmethvin. I agree on all counts.

entries/jQuery.when.xml
```diff
@@ -12,10 +12,15 @@
...
+  alert( err1.status );
 });
     </code></pre>
+    <p>In the previous example, <code>res1</code> and <code>res2</code> are arrays containing three values: <code>data</code>, <code>textStatus</code>, and <code>jqXHR</code>. It is adviced to always provide a callback to handle errors through the use of <code>deferred.fail()</code> or the second parameter of <code>deferred.then()</code>.</p>
+    <p>When a single <a href="/Types/#Deferred">Deferred</a>, <a href="/Types/#Promise">Promise</a>, or Promise-compatible object is passed to <code>jQuery.when()</code>, an argument for each of the returned value is provided to the <code>doneCallback</code> instead of a single argument containing an array of values. It is suggested not to rely on this behavior because it is considered an anti-pattern and it will change in the upcoming versions of jQuery, starting with version 3.</p>
```
gibson042: This is worded a little too strongly. What we really have are two distinct signatures (single-argument and multi-argument), each with distinct behavior ("cast"/"get as deferred"/etc. and "join", respectively) that will be well-tested as of jQuery 3.0.

mgol:
@AurelioDeRosa This PR is pretty old; will you be able to apply updates based on feedback?

AurelioDeRosa:
Is this PR still relevant or should I just close it?

mgol:
@AurelioDeRosa Since you opened this PR the `jQuery.when` page got only one small update in commit 3ab6720bdbeefa64d34bdb7d79bc03468788f49f. So I think it's still very much relevant (and rebasing over current master shouldn't be hard).

mgol:
@AurelioDeRosa what's the state of the PR? Will you have time to finish it?

Mansalol (approved):

Issue #921, opened 2016-05-05
Document jQuery.when changes in 3.0


gibson042:
Ref https://github.com/jquery/jquery/commit/356a3bccb0e7468a2c8ce7d8c9c6cd0c5d436b8b

PR #922, opened 2016-05-05
jQuery.when: Document changes in 3.0

gibson042:
Ref https://github.com/jquery/jquery/commit/356a3bccb0e7468a2c8ce7d8c9c6cd0c5d436b8b
Fixes gh-921


entries/jQuery.when.xml
```diff
@@ -3,60 +3,69 @@
...
+  <signature>
+    <added>1.5</added>
+    <argument name="values" type="Anything">
+    <desc>Two or values to resolve or encapsulate with Deferred objects for <code>Promise.all</code>-like fulfillment aggregation.</desc>
     </argument>
   </signature>
   <desc>Provides a way to execute callback functions based on one or more objects, usually <a href="/category/deferred-object/">Deferred</a> objects that represent asynchronous events.</desc>
   <longdesc>
-    <p>If a single Deferred is passed to <code>jQuery.when()</code>, its Promise object (a subset of the Deferred methods) is returned by the method. Additional methods of the Promise object can be called to attach callbacks, such as <a href="/deferred.then/"><code>deferred.then</code></a>. When the Deferred is resolved or rejected, usually by the code that created the Deferred originally, the appropriate callbacks will be called. For example, the jqXHR object returned by <code>jQuery.ajax()</code> is a Promise-compatible object and can be used this way:</p>
+    <p>If no arguments are provided, <code>jQuery.when()</code> will return a resolved Promise.</p>
```
dmethvin: I wonder if we need to create some other term than `Promise` since what's returned here is a `Deferred.promise` which is consumed a bit differently and has methods like `.done`, etc. The `types.xml` file does use `Promise` for this but now that we consume ES6 Promise it's ambiguous.

dmethvin:
Other than the question about nomenclature, LGTM!

entries/jQuery.when.xml
AurelioDeRosa: I agree with you @dmethvin. What if we start using the term jQuery Promise for our type, as we do for jqhxr, and Promise for ES6 Promises?

dmethvin: Maybe `jQuery-Promise` to make it clear it's a single term? What do you think @gibson042 ?

gibson042: I'm in favor of a jQuery-specific term, but would like to keep that out of this issue.

dmethvin: Yep no problem and I see you opened another issue. That lets this one continue on and we can return to figure out a good name.

Issue #796, opened 2015-08-10
`One` docs misleading


benjamingr:
The example for `.one` somewhat implies that `one` disables other event listeners added for the same event which is not the case.

Clarifying it could be nice.

AurelioDeRosa:
Hi @benjamingr. Can you clarify which example you are referring to and what kind of additional explanation you'd like us to add?

benjamingr:
I have a friend who started learning jQuery. He understood `one` as "it executes the event and removes all other event handlers for the event" rather than "it executes the event handler and removes it from the event" (which is what I told him it does and he refused to believe until I showed him in code).

I think the docs could be a little clearer on what it does :)

AurelioDeRosa:
@benjamingr are you interested in submitting a PR for this issue? We love contributions.

benjamingr:
I've talked to the friend and he agreed to work on a PR. I'll ask him to comment here about what caused the confusion and walk him through the process (I've been somewhat involved in jQuery before, but not the docs yet :)).

PR #923, opened 2016-05-05
Clarifying purpose of one()

ConnorCartwright:
Hey @AurelioDeRosa 

Regarding [issue #796](https://github.com/jquery/api.jquery.com/issues/796) I have changed the wording slightly and added in a code example using multiple event types to show that the handler can be called once for each event type.

### DIFF
--- a/entries/one.xml
+++ b/entries/one.xml
@@ -44,7 +44,7 @@
     </argument>
   </signature>
   <longdesc>
-    <p>The  <code>.one()</code> method is identical to  <code>.on()</code>, except that the handler is unbound after its first invocation. For example:</p>
+    <p>The  <code>.one()</code> method is identical to  <code>.on()</code>, except that the handler for a given element and event type is unbound after its first invocation. For example:</p>
     <pre><code>
 $( "#foo" ).one( "click", function() {
   alert( "This will be displayed only once." );
--- a/entries/one.xml
+++ b/entries/one.xml
@@ -61,7 +61,7 @@ $( "#foo" ).on( "click", function( event ) {
     <p>If the first argument contains more than one space-separated event types, the event handler is called <em>once for each event type</em>.</p>
     <pre><code>
 $( "#foo" ).one( "click mouseover", function() {
-  alert( "This could be displayed twice." );
+  alert( "The " + event.type + " event happened!" );
 });
     </code></pre>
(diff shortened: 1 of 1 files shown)


entries/one.xml
```diff
@@ -59,6 +59,12 @@ $( "#foo" ).on( "click", function( event ) {
     </code></pre>
     <p>In other words, explicitly calling <code>.off()</code> from within a regularly-bound handler has exactly the same effect.</p>
     <p>If the first argument contains more than one space-separated event types, the event handler is called <em>once for each event type</em>.</p>
+    <pre><code>
+$( "#foo" ).one( "click mouseover", function() {
+  alert( "This could be displayed twice." );
```
dmethvin: How about this?
```js
$( "#foo" ).one( "click mouseover", function( event ) {
  alert( "The first " + event.type + " event happened." );
});
```
That way it's clear it happens once for each event, although your text clarifications do that too.

entries/one.xml
```diff
@@ -59,6 +59,12 @@ $( "#foo" ).on( "click", function( event ) {
...
+});
+    </code></pre>
+    <p>In the example above the alert could be displayed twice due to the <i>two</i> event types (<code>click</code> and <code>mouseover</code>).</p>
```
AurelioDeRosa: Use `em` instead of `i`. They have a [slight different meaning](http://html5doctor.com/i-b-em-strong-element/).

entries/one.xml
ConnorCartwright: Hey @dmethvin 

I like this, I removed the 'first' as both could be triggered but I think displaying the event type as suggested is nicer!

entries/one.xml
ConnorCartwright: Hey @AurelioDeRosa 

I've updated this now, thank you for the feedback!

AurelioDeRosa:
LGTM, merging into the repository.

@jquery/content please join me in congratulating @ConnorCartwright on its first merged PR.

dmethvin:
:pizza: :star2:

kswedberg:
Woohoo! Congrats, @ConnorCartwright !

ConnorCartwright:
@AurelioDeRosa @dmethvin @kswedberg 

Haha, thanks guys! The first of many 😄

Issue #919, opened 2016-04-26
jQuery.ajaxSettings.traditional


dmethvin:
In http://api.jquery.com/jQuery.param/ we say that looking at the `traditional` flag was removed in 1.8 but in reality it didn't happen until 3.0. The note there needs to be updated to say 3.0.

AurelioDeRosa:
So, I'd say that we should remove the note from master and add it in branch 3.

dmethvin:
SGTM!

PR #924, opened 2016-05-06
jQuery.param: Removed details about traditional flag

AurelioDeRosa:
Fixes gh-919


dmethvin:
LGTM, it just needs to be moved to the 3.0 branch and the existing wording is fine.

AurelioDeRosa:
I was thinking that but because we forked it so long ago, in branch v3 we still have the text.

dmethvin:
Oh true, although it still says 1.8 right?

AurelioDeRosa:
Oh, true. I think I'm going to rebase the branch today because it's now very old. Once done that, I'm going to submit a PR to address it.

AurelioDeRosa:
I've submitted PR #932 to address the last problem left with this PR, so it's now good to merge.

Issue #405, opened 2013-12-23
Give a proper example of originalEvent, discourage use of jQuery.event.props


dmethvin:
Docs at http://api.jquery.com/category/events/event-object/ say to use `originalEvent` and then show the unadvisable practice of pushing additional properties on jQuery.event.props to be copied to the event object. Copying lots of properties slows down each delivered event.

DavidAustin:
originalEvent needs to be properly documented.  The line "Certain events may have properties specific to them" is weak and doesn't lead the reader to conclude that originalEvent is useful.

Certainly examples should be given.

dmethvin:
Since this is tagged as help wanted, we'd be glad to consider a pull request. Everyone here is a volunteer.

mgol:
See https://github.com/jquery/jquery/issues/3112. As @silverwind pointed out, we've now removed `jQuery.event.props` completely in https://github.com/jquery/jquery/commit/e61fccb9d736235b4b011f89cba6866bc0b8997d so this snippet is now not only not advised but will break in jQuery 3.0.0. We need to remove it ASAP and certainly before jQuery 3.0.0 comes out.

mgol:
I've added the `3.0.0` milestone.

silverwind:
Might be good to make a changelog entry too for the removal of `jQuery.event.props`.

mgol:
@silverwind What changelog do you have in mind?

mgol:
Ah, you probably mean the `Version` subsection. We have an in-progress branch at https://github.com/jquery/api.jquery.com/tree/v3.0.0-docs that will land when jQuery 3.0.0 comes but I see it's quite out of date (and doesn't even compile on my machine anymore). @AurelioDeRosa do you know what's the state of that?

silverwind:
I see you're not keeping a formal changelog. I meant the blog posts, like [this one](http://blog.jquery.com/2016/01/08/jquery-2-2-and-1-12-released/).

mgol:
@silverwind OK. :) We've already added the description to the [upgrade guide](https://jquery.com/upgrade-guide/3.0/#breaking-change-jquery-event-props-and-jquery-event-fixhooks-removed); there are no links to this page yet, we'll include one in the blog post.

PR #928, opened 2016-05-10
Event: Suggest event.originalEvent usage

silverwind:
Fixes: https://github.com/jquery/api.jquery.com/issues/405

cc: @mgol @dmethvin


silverwind:
Tweaked slightly.

mgol:
I like it!

dmethvin:
Yes! That also works for any version which is good, rather than being 3.0-specific.

mgol:
After 3.0 we'll need to add more props to the list but that may be done separately.

mgol:
Landed, thanks @silverwind!

PR #929, opened 2016-05-17
remove 4 self closing div tags

dennisbyrne:


AurelioDeRosa:
The jQuery function is able to create non-void elements even if the void version is provided. I agree that it isn't correct to do so and we should probably change it. Any comment from anyone before I merge?

dmethvin:
The API page for `jQuery` goes into more detail, but since this markup has an attribute it isn't handled through our shortcut `createElement` path and is instead passed to the browser's `innerHTML` property for parsing. IE<=8 `innerHTML` doesn't process self-closed *or* unclosed DIV tags properly, so those examples were broken for IE8 anyway. If we're over IE8 we can just remove the `/` with this PR.

The completely correct way do this would be `<div id='object1'></div>` since DIV and P officially can't be empty and must use the self-closing syntax. As long as all browsers since IE8 treat it right without the closing tag, I guess it's okay?

AurelioDeRosa:
Thank you for your comment @dmethvin. Based on it, I'd say let's make it 100% correct since we're at it. @dennisbyrne do you mind updating your PR?

dennisbyrne:
@AurelioDeRosa ... done.

PR #933, opened 2016-05-29
Explicit that .progress can take more than 2 arguments

tdelmas:
More conform with the .fail and .done doc


AurelioDeRosa:
Hi.

Thank you for the PR. The fix is good and I'll merge it as soon as you sign our CLA.

PR #939, opened 2016-06-16
Fix documentation for .val()

### LINKED ISSUE #828: Attr: selectMultipleWithNoValue.val() now returns [] instead of null
From https://github.com/jquery/jquery/issues/2562

We have this line...

> In the case of select elements, it returns null when no option is selected and an array containing the value of each selected option when there is at least one and it is possible to select more because the multiple attribute is present.

which could say something like this instead...

> When called on select elements that have no option selected, the `.val()` method returns `null`, or an empty array if the multiple attribute is present. If at least one option is selected, either the value or an array of values will be returned.

Also, one of the examples will need editing.

Where we have...

```js
var multipleValues = $( "#multiple" ).val() || [];
```
we can shorten it to...
```js
var multipleValues = $( "#multiple" ).val();
```

aried3r:
See:
https://github.com/jquery/api.jquery.com/issues/828
https://github.com/jquery/jquery/issues/2562
https://github.com/timmywil/jquery/commit/4319c2716f418b7dd3d5743c4b15cf5876fffed0

Fixes #828.

### DIFF
--- a/entries/val.xml
+++ b/entries/val.xml
@@ -13,6 +13,7 @@
     <longdesc>
       <p>The <code>.val()</code> method is primarily used to get the values of form elements such as <code>input</code>, <code>select</code> and <code>textarea</code>. When the first element in the collection is a <code>select-multiple</code> (i.e., a <code>select</code> element with the <code>multiple</code> attribute set), it returns an array containing the value of each selected option, or <code>null</code> if no options are selected. When called on an empty collection, it returns <code>undefined</code>.</p>
       <p>For selects and checkboxes, you can also use the <a href="/selected-selector/">:selected</a> and <a href="/checked-selector/">:checked</a> selectors to get at values, for example:</p>
+      <p><strong>jQuery 3</strong> changes the behavior of this method to return an empty array if a collection is a <code>select-multiple</code> and no options are selected.</p>
       <pre><code>
 // Get the value from a dropdown select
 $( "select.foo option:selected").val();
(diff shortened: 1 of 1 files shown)


entries/val.xml
```diff
@@ -115,8 +115,8 @@ $( "input" )
       <added>1.4</added>
       <argument name="function" type="Function">
         <argument name="index" type="Integer" />
-        <argument name="value" type="String" />     
-        <return type="String" />         
+        <argument name="value" type="String" />
+        <return type="String" />
```
aried3r: Should I leave the trailing whitespace as is or is it fine?

AurelioDeRosa: Whitespaces should be removed.

aried3r: Done

mgol:
LGTM.

mgol:
Sorry, not LGTM. We don't have versioned docs, every page describes the API in all jQuery versions so we need to have all documented. See e.g. how [the `data` page](https://api.jquery.com/data/) describes 3.0 behavior changes.

Could you try doing sth like that here?

aried3r:
> Could you try doing sth like that here?

I didn't know how to handle the example properly because I couldn't find an example. Hope it's fine like this.

AurelioDeRosa:
I'd suggest to update the PR to have something like we have for [`delegate()`](http://api.jquery.com/delegate/):

```js
// jQuery 1.4.3+
$( elements ).delegate( selector, events, data, handler );
// jQuery 1.7+
$( elements ).on( events, selector, data, handler );
```

aried3r:
How about
```JavaScript
// jQuery <3.0
var multipleValues = $( "#multiple" ).val() || [];
// jQuery 3.0+
var multipleValues = $( "#multiple" ).val()
```

`jQuery 1.0+` isn't quite correct, since that would include 3 as well. What do you think?

AurelioDeRosa:
Yeah, it looks good to me.

kswedberg:
Closed by 5689f1ade9bcbdb9bb9507f44ccf60dba167c2b8

PR #941, opened 2016-06-23
Update $.ajax() etc docs to not imply jqXHR.always() was in 1.5

laughinghan:
It wasn't: http://jsbin.com/sitoqem/edit?js,console

Other Promise/Deferred methods like `.done()` and `.fail()` are already available in 1.5.x, but `.always()` isn't added until jQuery 1.6: https://api.jquery.com/deferred.always/

https://api.jquery.com/jquery.getJSON/#jqxhr-object
https://api.jquery.com/jquery.get/#jqxhr-object
https://api.jquery.com/jquery.post/#jqxhr-object
http://api.jquery.com/jQuery.ajax/#jqXHR


mgol:
Landed, thanks! And sorry for waiting so long.

PR #946, opened 2016-07-06
[Wait for jQuery 3.1.0 release] jQuery.readyException: Add the page, add the 3.1 category

### LINKED ISSUE #942: Document jQuery.readyException
See https://github.com/jquery/jquery/pull/3210. `jQuery.readyException` is a callback that's fired when the function wrapped in `jQuery.ready` throws an error. By default it re-throws the error in a timeout so that it's not silenced but visible in the user console.

mgol:
Fixes #942

### DIFF
--- a/categories.xml
+++ b/categories.xml
@@ -422,6 +422,12 @@ var files = event.originalEvent.dataTransfer.files;
         <hr/>
       ]]></desc>
     </category>
+    <category name="Version 3.1" slug="3.1">
+      <desc><![CDATA[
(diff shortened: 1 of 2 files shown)


mgol:
That's what it looks like:
screen shot 2016-07-06 at 11 35 59

dmethvin:
LGTM

mgol:
I added a warning to the topic to not land it before the jQuery 3.1.0 release (that should happen this week). LGTMs are welcome, though. ;) (thanks @dmethvin!)

Issue #945, opened 2016-07-03
delegate() is not listed as deprecated


teo1978:
`.delegate()` isn't listed here:
http://api.jquery.com/category/deprecated/

nor is there any notice here: http://api.jquery.com/delegate/ saying that it is officially deprecated. It only says:
> As of jQuery 1.7, .delegate() has been superseded by the .on() method.

However, a search for the word "deprecated" in the page will find nothing.

Yet, if you use the jQuery 3 Migrate plugin, you will get warnings if you use `delegate()` that it is deprecated.

markelog:
Nice catch! Would you like to send us a pr for that?

PR #948, opened 2016-07-10
delegate: Document the method was deprecated in jQuery 3

AurelioDeRosa:
Fixes gh-945

### DIFF
--- a/entries/bind.xml
+++ b/entries/bind.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<entry type="method" name="bind" return="jQuery">
+<entry type="method" name="bind" return="jQuery" deprecated="3.0">
   <title>.bind()</title>
   <signature>
     <added>1.0</added>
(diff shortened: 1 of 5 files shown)


mgol:
What about undekegate, bind & unbind?

AurelioDeRosa:
I assumed it was missing from delegate only because of the issue. I'm going to update the PR.

entries/bind.xml
```diff
@@ -34,7 +34,7 @@
...
+    <p>As of jQuery 3.0, <code>.bind()</code> has been deprecated. It was superseded by the <a href="/on/"><code>.on()</code></a> method for attaching event handlers to a document since jQuery 1.7, so its use was already discouraged. For earlier versions, the <code>.bind()</code> method is used for attaching an event handler directly to elements. Handlers are attached to the currently selected elements in the jQuery object, so those elements <em>must exist</em> at the point the call to <code>.bind()</code> occurs. For more flexible event binding, see the discussion of event delegation in <a href="/on/"><code>.on()</code></a> or <a href="/delegate/"><code>.delegate()</code></a>.</p>
```
mgol: I wouldn't direct people to docs for deprecated APIs so I'd remove the reference to https://api.jquery.com/delegate/

entries/undelegate.xml
```diff
@@ -43,7 +43,7 @@
     </argument>
   </signature>
   <longdesc>
-    <p>The <code>.undelegate()</code> method is a way of removing event handlers that have been bound using <a href="/delegate/"><code>.delegate()</code></a>. <strong>As of jQuery 1.7</strong>, the <a href="/on/"><code>.on()</code></a> and <a href="/off/"><code>.off()</code></a> methods are preferred for attaching and removing event handlers.</p>
+    <p>The <code>.undelegate()</code> method is a way of removing event handlers that have been bound using <a href="/delegate/"><code>.delegate()</code></a>. As of jQuery 3.0, <code>.undelegate()</code> has been deprecated. It was superseded by the <a href="/off/"><code>.off()</code></a> method since jQuery 1.7, so its use was already discouraged.</p>
```
mgol: The deprecation message should appear first, before any description of the API. See e.g. https://api.jquery.com/live/. This applies to the `unbind` page as well.

mgol:
A few minor remarks, otherwise it's good.

entries/bind.xml
AurelioDeRosa: The method is available anyway, so removing the link only add frustration in having to open a new tab, searching for the method, and then read the doc. We've already stated the correct method to use is `.on()`.

entries/undelegate.xml
AurelioDeRosa: Done.

entries/bind.xml
mgol: But why would anyone search for docs for this method? Most likely because they use `.delegate()` which can't be said about all the readers of the `.bind()` docs.

I think we shouldn't link to pages about deprecated APIs, they should serve first as a warning to not use the API and second as docs for the API for people that *already use* it in their code. If there is useful info about event delegation there that's not available on the `.on()` page it should just be moved to the `.on()` page so that it's not required to look at pages for deprecated APIs to learn something generally useful.

(to clarify, I wanted to have the `.delegate()` mention removed completely, not just the link to the page)

mgol: @AurelioDeRosa Hey, do you have any remarks to my comment? For me this is the only thing left before this can be landed.

mgol: @AurelioDeRosa Ping? We need this PR to land to unbreak the build.

@jquery/core any opinion on my comments? I need someone to agree with either of us so that we can land it.

mgol: @AurelioDeRosa Ping? We need this PR to land to unbreak the build.

@jquery/core @jquery/content any opinion on my comments? I need someone to agree with either of us so that we can land it.

dmethvin: The main reason someone might search specifically for `.delegate()` would be to figure out how to convert a call to `.on()`. That wouldn't apply for this reference. Any explanation of delegated events in `.delegate()` should be duplicated in `.on()`, I think it already should be there but if not we can open a new ticket to fix it. 

If #971 can be addressed adequately there (e.g. via a watermark) there wouldn't be any doubt about the method being deprecated and/or removed. As I mentioned in #972 I think there might be good reasons to keep, for example, the `click`-specific information in the `.click()` method docs, but that doesn't apply here either.

mgol: @dmethvin As I understand you, you agree with me it's best not to link to `.delegate()` here? I'll go ahead & remove the link and land it then so that we have an unbroken build.

PR #955, opened 2016-07-14
val(): Update select multiple no selected statement for 3.0. Fixes #828

### LINKED ISSUE #828: Attr: selectMultipleWithNoValue.val() now returns [] instead of null

kswedberg:
Added a statement about the difference between >=3.0 and <3.0 when calling .val() on &lt;select multiple> with no options selected.


mgol:
Have you seen #939? What do you think about code example changes there?

kswedberg:
Oh, bummer! I didn't see that one. The code example change there looks fine to me. Maybe we could merge that one first. Then merge mine, keeping that code change but using my content change?

mgol:
Sounds good. Would you do that?

-- 
Michał Gołębiowski

kswedberg:
sure.

gibson042:
LGTM.

entries/val.xml
```diff
@@ -115,17 +116,17 @@ $( "input" )
...
       <p>Setting values using this method (or using the native <code>value</code> property) does not cause the dispatch of the <code>change</code> event. For this reason, the relevant event handlers will not be executed. If you want to execute them, you should call <code>.trigger( "change" )</code> after setting the value.</p>
-      <p>The <code>.val()</code> method allows us to set the value by passing in a function. As of jQuery 1.4, the function is passed two arguments, the current element's index and its current value: </p>
+      <p>The <code>.val()</code> method allows settting the value by passing in a function. As of jQuery 1.4, the function is passed two arguments, the current element's index and its current value: </p>
```
AurelioDeRosa: "settting" should be "setting"

kswedberg: good catch. thanks!

Issue #956, opened 2016-07-17
Update removed category for v3.0


Mottie:
The [removed](http://api.jquery.com/category/removed/) category page needs to be updated for jQuery v3.0+. In most cases the "removed: 3.0" in the header of each page is also missing.

I don't have an extensive list of removed items, but so far I've found:

| Item                                           | Deprecated | Removed |
|------------------------------------------------|:----------:|:-------:|
| [`.andSelf()`](http://api.jquery.com/andSelf/) | 1.8        | 3.0     |
| [`.error`](http://api.jquery.com/error/)       | 1.8        | 3.0     |
| [`.load`](http://api.jquery.com/load-event/)   | 1.8        | 3.0     |
| [`.size()`](http://api.jquery.com/size/)       | 1.8        | 3.0     |
| [`.unload`](http://api.jquery.com/unload/)     | 1.8        | 3.0     |
| [`.context`](http://api.jquery.com/context/)   | 1.10       | 3.0     |

PR #957, opened 2016-07-17
Updated methods removed in jQuery 3

AurelioDeRosa:
Fixes gh-956


mgol:
The OP also reported a `removed` category entry missing for `.context`.

AurelioDeRosa:
PR Updated.

mgol:
LGTM.

It'd be good to check if we don't have any other entries with `removed="sth"` that don't belong to the `removed` category but that's another story.

PR #960, opened 2016-08-01
Speed of selectors

vitaliylag:
Selectors of the form tag#id.class are not fast selectors and can not be processed very quickly. They are working with O(n) so I fixed confusion in the documentation.


AurelioDeRosa:
Why are you saying that ID and class selectors aren't very fast? jQuery takes advantage of native methods when using those selectors which makes it very fast.

vitaliylag:
I have tested it on jQuery 1.11.0 and 3.1.0. When I adding 100k listeners with '.class' selector every calling trigger takes more than 6 second!

Well, somebody can say that 100k is too much but 1000 listeners is usual case and it lags for 60 msec every click. On the mobile phonse it can lag about 1 second. It is very slow.

ID selectors work faster but only 5 times faster and still very slow.

Do you need the tests?

vitaliylag:
When I reimplamented standart $.fn.on it began to work 1200 times faster with trigger on class selector with 100k listeners and 2 times faster on adding listeners (and this is using original $.fn.on for compatibility with $.fn.trigger). My implamentation break calling orders but I don't need it at all.

Event with 1000 listeners my implamentation works 30 times faster, 9 times faster with 100 listeners and 2 times faster with 10 listeners.

dmethvin:
Those are very unrealistic tests. Most elements have only one or two event listeners on them. Thousands on a single element would be a sign of something seriously wrong with the way the developer has designed their application. Just the extra memory we use to track events would become significant at that point.

vitaliylag:
I don't think 1000 listeners on one elements is bad. I have one-page application with autogenerated content and it is normal.

Memory is not issue. When I reimplemented it 1000 listeners use about 50 KB memory. Issue is your bad implementation. And this PR about wrong documentation. I don't mind bad implementation but it should be documentated properly.

vitaliylag:
@dmethvin, I checked jQuery, it doesn't use much memory too. May be you mean listeners **without filtering**? I'm saying about filtering events by selectors, not just events.

AurelioDeRosa:
@vitaliylag if you think there is something wrong with the implementation, you should open an issue  and perhaps submit a PR to fix it. I agree with @dmethvin that adding 1000 listeners to an element sounds like a crazy idea and that a test where you add 100k listeners to an element doesn't match any real use-case scenario.

vitaliylag:
«1000 listeners sounds like a crazy idea» — it's not crazy idea, in my one-page application I have 200 listeners and it's not much.

vitaliylag:
@AurelioDeRosa, I have opened issue with description of fast algorithm keeping calling order https://github.com/jquery/api.jquery.com/issues/961, but that doesn't mean that documentation shouldn't be improved.

PR #968, opened 2016-08-10
Filter: Explain behavior on text and comment nodes

dmethvin:
This enshrines the behavior that was present before 3.0 but undocumented. See https://github.com/jquery/jquery/issues/3226 .


gibson042:
LGTM.

PR #969, opened 2016-08-10
Deprecate jQuery.fx.interval as of 3.0

dmethvin:
Ref https://github.com/jquery/jquery/issues/3154

### DIFF
--- a/entries/jQuery.fx.interval.xml
+++ b/entries/jQuery.fx.interval.xml
@@ -1,12 +1,13 @@
 <?xml version="1.0"?>
-<entry type="property" name="jQuery.fx.interval" return="Number">
+<entry type="property" name="jQuery.fx.interval" return="Number" deprecated="3.0">
   <title>jQuery.fx.interval</title>
   <desc>The rate (in milliseconds) at which animations fire.</desc>
   <signature>
     <added>1.4.3</added>
   </signature>
   <longdesc>
-    <p>This property can be manipulated to adjust the number of frames per second at which animations will run. The default is 13 milliseconds. Making this a lower number could make the animations run smoother in faster browsers (such as Chrome) but there may be performance and CPU implications of doing so.</p>
+    <p>This property is deprecated as of version 3.0, and has no effect in browsers that support the <a href="http://caniuse.com/#feat=requestanimationframe"><code>requestAnimationFrame</code></a> method.</p>
+    <p>On browsers that do not support <code>requestAnimationFrame</code>, this property can be changed to adjust the interval at which animations will run. The default is 13 milliseconds.</p>
     <p>Since jQuery uses one global interval, no animation should be running or all animations should stop for the change of this property to take effect.</p>
   </longdesc>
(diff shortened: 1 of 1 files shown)


AurelioDeRosa:
LGTM

entries/jQuery.fx.interval.xml
```diff
@@ -1,12 +1,13 @@
 <?xml version="1.0"?>
-<entry type="property" name="jQuery.fx.interval" return="Number">
+<entry type="property" name="jQuery.fx.interval" return="Number" deprecated="3.0">
```
mgol: It's not deprecated in 3.0 but will be in 3.2. If there's a category missing, it should be added.

Do we need to look into things currently tagged as `deprecated="3.0"`? May there be more things to correct there?

entries/jQuery.fx.interval.xml
```diff
@@ -34,4 +35,5 @@ $( "input" ).click(function() {
   <category slug="effects/custom-effects"/>
   <category slug="properties/global-jquery-object-properties"/>
   <category slug="version/1.4.3"/>
+  <category slug="deprecated/deprecated-3.0"/>
```
mgol: Same as above, `deprecated/deprecated-3.2`.

entries/jQuery.fx.interval.xml
dmethvin: Is there anything marked deprecated-3.0? It looks like all the deprecations from the upgrade guide are missing.

It might be good to make this particular one effective with 3.0 since we started using `rAF` then. For the others I agree, 3.2 would be better.

mgol: I see. I guess this makes sense to target 3.0 then, our Core ticket still mentions 3.2.0, though, IIRC.

I'll see what's up with other deprecations.

Krinkle:
This failed the build. The 3.0 deprecation category doesn't exist yet.

One of the currently open pull requests creates this category, though. ( https://github.com/jquery/api.jquery.com/pull/948)

```
Warning: post jQuery.fx.interval has a category term slug of 'deprecated/deprecated-3.0', but no such term exists. Use --force to continue.
```

mgol:
@Krinkle Thanks for the update. I pinged Aurelio & a couple of teams, I'll try to get this landed soon.

dmethvin:
Ack! Sorry, my fault, I am running on Windows and have never been able to get things running properly on a local build. Can someone who has a working local setup try #948 and then land it?

Krinkle:
The site deployment infrastructure was broken for several weeks while so even the commit before this one wasn't deployed yet. I've manually triggered a build for e55cf6302af7ee9fe849d062998ebda0510ad43f (1.12.11) just now, so that this no longer blocks other changes waiting for deployment.

mgol:
I landed #948, deployment now works fine for the Core API site.

dmethvin:
:fireworks: :dancers: :fireworks:

PR #992, opened 2016-10-19
deferred.catch: create the page

### LINKED ISSUE #951: Document deferred.catch
We should have a page at https://api.jquery.com/deferred.catch/ that documents `.catch(fn)` as an alias of `.then(null, fn)`. It has been available since 3.0.

mgol:
Fixes #951


mgol:
screen shot 2016-10-19 at 14 35 27

Issue #997, opened 2016-11-13
Empty-context dimensions getters return undefined as of 2.2.0


gibson042:
Ported from jquery/jquery/issues/3396 .
See jquery/jquery/issues/2319 and jquery/jquery/pull/2701 for explanation.

PR #998, opened 2016-11-13
Dimensions: Document empty-context getters

gibson042:
Fixes gh-997

### DIFF
--- a/entries/innerHeight.xml
+++ b/entries/innerHeight.xml
@@ -9,7 +9,7 @@
   </signature>
   <desc>Get the current computed height for the first element in the set of matched elements, including padding but not border.</desc>
   <longdesc>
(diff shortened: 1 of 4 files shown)


entries/innerHeight.xml
```diff
@@ -9,7 +9,7 @@
   </signature>
   <desc>Get the current computed height for the first element in the set of matched elements, including padding but not border.</desc>
   <longdesc>
-    <p>This method returns the height of the element, including top and bottom padding, in pixels.</p>
+    <p>This method returns the height of the element, including top and bottom padding, in pixels. If called on an empty set of elements, returns <code>undefined</code> (<code>null</code> before jQuery 2.2.0).</p>
```
mgol: `null` before jQuery 3.0, not 2.2.0. Same below.

gibson042: Hmm, https://github.com/jquery/jquery/blame/master/src/dimensions.js#L50 must be lying somehow (or I'm misreading it). But you're right. Fixed.

mgol:
LGTM, it can be landed as two commits.

Issue #999, opened 2016-11-23
Issue in Ajax Call


lgchintake:
I am using **jQuery v2.1.4.min.js** library.

I am calling ajax giving some parameters to it as like following:

**First Case**
```
var myData = '{"userId":"abc", "password":"abc??"}';
$.ajax({
	type: "POST",
	url: myUrl,
	data: myData,
	dataType: "json",
	success: function(serviceData) {
		console.log(JSON.stringify(serviceData));
	},
	error: function(err) {
		console.log("Error: "+err);
	}
});

```
I have given password having two question marks(??) so it is manipulating it automatically as 

{"userName":"abc", "password":"abcjQuery21406515378922229067_1479880736745?"}:

When I remove the `dataType: "json",` it works fine.

**Second Case**

When I write above code with ContentType as:

```
var myData = '{"userId":"abc", "password":"abc??"}';
$.ajax({
	type: "POST",
	url: myUrl,
        contentType: "application/json; charset=utf-8",
	data: myData,
	dataType: "json",
	success: function(serviceData) {
		console.log(JSON.stringify(serviceData));
	},
	error: function(err) {
		console.log("Error: "+err);
	}
});

```

It works fine. No issue.

I am using Java RestEasy Service for handling the ajax requests.

dmethvin:
This is a dup of gh-1799.

mgol:
@dmethvin There is no issue 1799 in this repo; where did you want to link?

lgchintake:
@**dmethvin** Where to find issue no. 1799 ?

dmethvin:
Sorry, it's here: https://github.com/jquery/jquery/issues/1799

Since this is the docs repo, I suppose this is a request to add something to the `jQuery.ajax` page about it? It doesn't look like it's currently mentioned.

seema0518:
I am using below code:
$("#submit").click(function(e) {
    var jsonData = {  
    "name":"Pavan Verma",
    "email":"myname@gmail.com",
    "company_name":"My Company",
    "company_website":"",
    "country_code":"+91",
    "phone":"9876543210",
    "skype":"",
    "message":"test message1\ntest message2\ntest message3",
    "solutions":[],
    "contact_methods":[],
    "other_contact_methods":null,
    "page_url":"https://test.orgzit.com/"
};
$.ajax({
    type: 'POST',
    url: 'https://www.orgzit.com/api/1/contactform/',
    data: jsonData,
[...]

Issue #878, opened 2016-01-25
Ajax: add note on how to bypass query string alteration for JSON requests


timmywil:
See https://github.com/jquery/jquery/issues/1799

JSON requests will sometimes get promoted to JSONP if deemed cross-browser (we already document that part). We add a callback name to the query string for JSONP requests (we document that too). The part that isn't documented is that in rare cases, the data for a json request contains a question mark, and so the callback name may be injected into a weird part of the string. Users can work around this by setting `jsonp` to false.

PR #1005, opened 2016-12-19
ajax: Specified that "?" in data is replaced

AurelioDeRosa:
Fixes gh-999


dmethvin:
I think this is too broad a statement, see this http://jsbin.com/dicigovevo/edit?html,console and also #878 which is a bit of a dup. There's some lengthy discussion in https://bugs.jquery.com/ticket/8417 .

The essence of the problem is that when users send their own plain string (instead of an object) as `data` they must ensure that it is either formatted properly as form-url-encoded data or else specify a `contentType` that says how the data is encoded.  It is not possible for form-url-encoded data to contain `??` by the definition of the format. 

I kept trying to explain it all and I ended up with something like this, which is probably 3 paragraphs.

> Data to be sent to the server. If the HTTP method is one that cannot have an entity body, such as GET, the `data` is appended to the URL. When `data` is an object, jQuery generates the data string from the object's key/value pairs unless the `processData` option is set to `false`. For example, `{ a: "bc", d: "e,f" }` is converted to the string `"a=bc&d=e%2Cf"`. If the value is an array, jQuery serializes multiple values with same key based on the value of the `traditional` setting (described below). For example, `{ a: [1,2] }` becomes the string `"a%5B%5D=1&a%5B%5D=2"` with the default `traditional: false` setting. When `data` is passed as a string it should *already be encoded* using the correct encoding for `contentType`, which by default is `application/x-www-form-urlencoded`. If the string contains a double question mark (`??`) it is replaced with a value generated by jQuery that is unique for each copy of the library on the page (e.g. jQuery21406515378922229067_1479880736745).

mgol:
@dmethvin it seems that your version is pretty different to the one in this PR. Could you report it as a PR? We'd be able to close this one then (I'm trying to clear up stale PRs).

mgol:
@dmethvin ping

AurelioDeRosa:
Agree with @mgol. @dmethvin's text is much more complete, so happy to close mine in favor of a PR fro him.

mgol:
@dmethvin Ping about https://github.com/jquery/api.jquery.com/pull/1005#issuecomment-273006794

mgol:
I submitted @dmethvin's text in #1171 so let's close this one.

Issue #989, opened 2016-10-09
.outerHeight( function(index, height) ) doesn't specify the function prototype


tdelmas:
https://api.jquery.com/outerheight/#outerHeight-functionindex--height specify only
`Type: Function()`
https://api.jquery.com/innerheight/#innerHeight-function is more specific:
`Type: Function( Integer index, Number height ) => String or Number`
I think for outerHeight it should be the same?

AurelioDeRosa:
Thank you for opening this issue. I'll fix this ASAP.

PR #1009, opened 2016-12-19
outerHeight(): Specified type of passed arguments to callback

AurelioDeRosa:
Fixes gh-989

### DIFF
--- a/entries/outerWidth.xml
+++ b/entries/outerWidth.xml
@@ -56,7 +56,13 @@ $( "p:last" ).text(
   </signature>
   <signature>
     <added>1.8.0</added>
-    <argument name="function(index, width)" type="Function">
+    <argument name="function" type="Function">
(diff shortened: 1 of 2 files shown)


mgol:
This looks good but the same changes should be applied to `outerWidth`. The `inner*` variants already look fine.

AurelioDeRosa:
@mgol I've updated the PR.

entries/outerWidth.xml
```diff
@@ -56,7 +56,13 @@ $( "p:last" ).text(
   </signature>
   <signature>
     <added>1.8.0</added>
-    <argument name="function(index, width)" type="Function">
+    <argument name="function" type="Function">
+      <argument name="index" type="Integer" />
+      <argument name="height" type="Number" />
```
mgol: `width`, not `height` :)

PR #1010, opened 2016-12-20
load(): Specified that load event alias is deprecated

### LINKED ISSUE #976: The .load() page still links to the old event alias
The page at https://api.jquery.com/load/ still says:

> Note: The event handling suite also has a method named `.load()`. jQuery determines which method to fire based on the set of arguments passed to it.

This is no longer true as of jQuery 3.0, the second method was removed and there is no logic to detect which method to invoke. We should either remove this note or make it say it **used to** behave like that but it no longer does.

AurelioDeRosa:
Fixes #976

### DIFF
--- a/entries/load.xml
+++ b/entries/load.xml
@@ -21,7 +21,7 @@
   <desc>Load data from the server and place the returned HTML into the matched element.</desc>
   <longdesc>
     <div class="warning">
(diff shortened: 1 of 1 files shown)


mgol:
The code changes look good but the commit message is incorrect - the alias is now no longer just deprecated but removed.

AurelioDeRosa:
Thank you for the review. I'm going to change the commit message when rebasing.

PR #1018, opened 2017-01-15
serializeArray: Document Behavior for Value-less Input Elements

jugglinmike:
Formally define jQuery's behavior when `<input>` elements which specify a
`name` property but no corresponding `value`. This behavior should be considered
stable because it is implicitly asserted in jQuery's unit test suite through
tests for the `serialize` method [1].

[1] https://github.com/jquery/jquery/blob/efdb8a46e4213dcf69f792c42c234c6b112ba471/test/unit/serialize.js#L125-L127

The expected behavior here is a little ambiguous because it could also be valid to simply skip such elements in the serialization process. That is [the Cheerio project](https://github.com/cheeriojs/cheerio/)'s current behavior, but before [altering it to match jQuery's current behavior](https://github.com/cheeriojs/cheerio/pull/842), I'd like to make the behavior explicit.

### DIFF
--- a/entries/serializeArray.xml
+++ b/entries/serializeArray.xml
@@ -28,7 +28,7 @@
   &lt;/div&gt;
 &lt;/form&gt;
     </code></pre>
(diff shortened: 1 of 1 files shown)


jugglinmike:
Thanks, Dave! Is there anything more I can do to get this patch landed?

jugglinmike:
Thanks, Karl!

Issue #1019, opened 2017-01-22
Mandatory key argument to jqXHR#getResponseHeader is not documented


gibson042:
Ref https://github.com/jquery/jquery/issues/3508

jqXHR objects expose a `getResponseHeader` method which accepts a single mandatory string `key`, but that argument is not documented. jqXHR itself is buried at http://api.jquery.com/jQuery.ajax/#jqXHR , but this particular issue can be fixed without correcting that deeper issue.

PR #1020, opened 2017-01-22
Improve jqXHR documentation

gibson042:
Fixes gh-1019

### DIFF
--- a/entries/jQuery.ajax.xml
+++ b/entries/jQuery.ajax.xml
@@ -281,16 +281,16 @@ jqxhr.always(function() {
         <code>readyState</code>
       </li>
       <li>
+        <code>responseXML</code> and/or <code>responseText</code> when the underlying request responded with xml and/or text, respectively
+      </li>
+      <li>
         <code>status</code>
       </li>
       <li>
         <code>statusText</code>
       </li>
       <li>
-        <code>responseXML</code> and/or <code>responseText</code> when the underlying request responded with xml and/or text, respectively
-      </li>
-      <li>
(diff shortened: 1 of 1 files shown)


entries/jQuery.ajax.xml
```diff
@@ -281,24 +281,31 @@ jqxhr.always(function() {
...
         <code>statusText</code>
       </li>
-      <li><code>responseXML</code> and/or <code>responseText</code> when the underlying request responded with xml and/or text, respectively</li>
-      <li><code>setRequestHeader(name, value)</code> which departs from the standard by replacing the old value with the new one rather than concatenating the new value to the old one</li>
       <li>
-        <code>getAllResponseHeaders()</code>
+        <code>abort( [ statusText ] )</code>
+      </li>
+      <li>
+        <code>getAllResponseHeaders()</code> as a string
```
dmethvin: Is it worth adding "as a string" here? The others don't talk about return types in general.

gibson042: Absent prior knowledge, I'd expect `getAllResponseHeaders()` to return a map. This clarifies that to not be the case.

AurelioDeRosa:
Should I keep all the commits here or should I fix up them?

gibson042:
They'll always be distinct here; I'm not opposed to a landing them on master as a single unit.

AurelioDeRosa:
Yeah, my question was basically if there was any preference on how to merge them in master.

gibson042:
I personally prefer the squashed representation, and that's also our standard in Core and QUnit.

mgol:
Is anything blocking landing this?

kswedberg:
@mgol Not that I'm aware of. LGTM

PR #983, opened 2016-09-20
Document jQuery.ready

Krinkle:
Also:
- Categorise `jQuery.holdReady` in events/document-loading,
  to match `ready()` and `jQuery.ready.promise()`.
- Categorise `jQuery.holdReady` in properties/global-jquery-object-properties,
  to match `jQuery.fx.off()` and `jQuery.ready.promise()`.

Fixes #205

PR #1026, opened 2017-03-16
jQuery.when: Add Thenable and Promise as acceptable types

### LINKED ISSUE #906: .when() documentation should say it takes Promises, not just Deferreds
The [documentation](https://api.jquery.com/jquery.when/) for $.when() says that it takes Deferreds. However, it later implies that it can take Deferreds or Promises:

> If a single argument is passed to jQuery.when() and it is not a Deferred or a Promise, ...

If so, then the documentation should be updated to say that $.when() takes Promises, not just Deferreds. Promises are more general than Deferreds.

In my testing, $.when() works with Promises, not just Deferreds.

Krinkle:
Technically replacing Deferred with Thenable would work, but mentioning Deferred and Promise directly should make this easier to understand for new and existing users.

Fixes #906.

### DIFF
--- a/pages/Types.html
+++ b/pages/Types.html
@@ -93,6 +93,7 @@ var x = "";
   <li class="toclevel-1"><a href="#jQuery"><span class="toctext">jQuery</span></a></li>
   <li class="toclevel-1"><a href="#XMLHttpRequest"><span class="toctext">XMLHttpRequest</span></a></li>
   <li class="toclevel-1"><a href="#jqXHR"><span class="toctext">jqXHR</span></a></li>
+  <li class="toclevel-1"><a href="#Thenable"><span class="toctext">Thenable</span></a></li>
   <li class="toclevel-1"><a href="#Deferred"><span class="toctext">Deferred Object</span></a></li>
   <li class="toclevel-1"><a href="#Promise"><span class="toctext">Promise Object</span></a></li>
   <li class="toclevel-1"><a href="#Callbacks"><span class="toctext">Callbacks Object</span></a></li>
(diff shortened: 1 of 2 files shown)


Krinkle:
This is a minimal change to unblock #983. A few other changes to consider separately:
* Add an optional second parameter that is the same, to indicate that it takes variadic arguments, similar to `append()`.
* Add a secondary signature with `Anything` and improve documentation on the behaviour of turning non-thenables into immediately-resolved deferreds.

pages/Types.html
```diff
@@ -657,6 +658,8 @@ <h2 id="XMLHttpRequest"> XMLHttpRequest </h2>
 <h2 id="jqXHR"> jqXHR </h2>
 <p>As of jQuery 1.5, the <a href="/jQuery.ajax/">$.ajax()</a> method returns the jqXHR object, which is a superset of the XMLHTTPRequest object. For more information, see the <a href="/jQuery.ajax/#jqXHR">jqXHR section of the $.ajax entry</a>
 </p>
+<h2 id="Thenable">Thenable</h2>
+<p>Any object with a <code>then</code> method that returns a <a href="#Promise">Promise</a>-compatible object.</p>
```
mgol: We should stick to the definition at https://promisesaplus.com/#point-7 i.e. the only assumption about `then` is that it's a function.

mgol: Maybe mentioning how it relates to promises somehow would clarify things but it should be obvious this is only an advisory remark and not part of the definition.

Krinkle: I've simplified the phrasing, leaving out the Promise mention for now.

pages/Types.html
```diff
@@ -657,6 +658,8 @@ <h2 id="XMLHttpRequest"> XMLHttpRequest </h2>
 <h2 id="jqXHR"> jqXHR </h2>
 <p>As of jQuery 1.5, the <a href="/jQuery.ajax/">$.ajax()</a> method returns the jqXHR object, which is a superset of the XMLHTTPRequest object. For more information, see the <a href="/jQuery.ajax/#jqXHR">jqXHR section of the $.ajax entry</a>
 </p>
+<h2 id="Thenable">Thenable</h2>
+<p>Any object that has a <code>then</code> method.</p>
```
mgol: Hmm, the spec at https://promisesaplus.com/#point-7 mentions functions with a `then` method as acceptable input as well.

That's a little weird one but technically we're still accepting those so maybe we should keep our definition identical to the Promises/A+ one?

Krinkle: @mgol I considered that but found it more confusing than useful. All functions are objects. So this isn't contradictory.

mgol: Hmm, OK. I guess it was mentioned as that's the only object that has a non-`"object"` `typeof`.

PR #1042, opened 2017-06-07
Remove delegated on example

FokkeZB:
This example doesn't work since the link will have opened before the event has bubbled up to body.


dmethvin:
Not sure what you mean. This works for me. http://jsbin.com/deramuvoni/1/edit?html,console,output

FokkeZB:
Ah man, sorry for that. As it turned out, another script was overriding the event handling causing this.

👮 Nothing to see here, move along...

PR #1044, opened 2017-06-12
fix typo

atefBB:


dmethvin:
That's not a typo. In English a soft "h" is preceded by "an" so " a hard problem" vs. "an honorable man".

atefBB:
Thx !

PR #1047, opened 2017-06-14
Changed typo behavior to behaviour

sagarvd01:


mgol:
That's not a typo. We're using American English.

PR #1049, opened 2017-06-21
add accepted param types for html method

MrSkinny:
Noticed that `.html()` accepts more than just an `htmlString` but the docs don't reflect this.

### DIFF
--- a/entries/html.xml
+++ b/entries/html.xml
@@ -68,8 +68,11 @@ $( "p" ).click(function() {
   <entry type="method" name="html" return="jQuery">
     <signature>
       <added>1.0</added>
-      <argument name="htmlString" type="htmlString">
-        <desc>A string of HTML to set as the content of each matched element.</desc>
+      <argument name="content">
+        <desc>A string of HTML, DOM element, or array of strings or elements to set as the content of each matched element.</desc>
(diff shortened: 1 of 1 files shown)


mgol:
What would be the expected output? Remember that the fact it doesn't crash doesn't mean it works and is supported. I don't think an object or array input would be expected here so it should stay as it is.

AurelioDeRosa:
Not talking for the OP, but I can imagine:

- jQuery object: same as calling $el.html()
- Array is a bit complicated because it depends on the content, so I won't be supportive of this.

mgol:
> jQuery object: same as calling $el.html()

I've looked at the code and it seems that for jQuery objects current code makes the following:
```
elem.html(elem2);
```
equivalent to:
```
elem.empty().append(elem2);
```
which means it affects the input element. I'm not sure if that's a good API and if - because of that - we should officially support it. What do you think, @jquery/core?

Adding support would also require adding unit tests for the feature to jQuery before it's added to the API page.

mgol:
At the today's Core meeting we agreed supporting elements & arrays as input to `.html()` doesn't make for a good API so even if it sort-of works we won't be documenting that (or supporting it officially).

PR #1054, opened 2017-07-05
Add missing signature

oriadam:
The signature of passing a settings object instead of URL string was added on jQuery 1.12 but not documented.
Added the signature and updated the cache related workarounds and examples to reflect the new signature.
The example with loading jquery-color was edited to support caching because browser caching is highly recommended when loading a plugin code.

### DIFF
--- a/entries/jQuery.getScript.xml
+++ b/entries/jQuery.getScript.xml
@@ -70,7 +70,7 @@ $( "div.log" ).ajaxError(function( e, jqxhr, settings, exception ) {
     </code></pre>
     <h4 id="caching-requests">Caching Responses</h4>
     <p>By default, <code>$.getScript()</code> sets the cache setting to <code>false</code>. This appends a timestamped query parameter to the request URL to ensure that the browser downloads the script each time it is requested. 
-      To override this feature you can use the settings object signature as of jQuery 1.12:
+      You can override this feature by using a settings object, as of jQuery 1.12:</p>
       <pre><code>
 $.getScript({
   url: "foo.js",
@@ -78,7 +78,7 @@ $.getScript({
 });
     </code></pre>
     
-    You can also override this feature by setting the cache property globally using <a href="/jquery.ajaxsetup/"><code>$.ajaxSetup()</code></a>:
+    <p>You can also override this feature by setting the cache property globally using <a href="/jquery.ajaxsetup/"><code>$.ajaxSetup()</code></a>:</p>
     <pre><code>
 $.ajaxSetup({
   cache: true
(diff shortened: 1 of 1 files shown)


AurelioDeRosa:
Hi @oriadam, thank you for your contribution. Could you update the examples in your PR to match our style guide? You can find the guide here: https://contribute.jquery.org/style-guide/js/

oriadam:
Hi @AurelioDeRosa be glad to :) is it ok now ?

mgol:
We don't seem to have a style guide for XML documents. The one that @AurelioDeRosa links to describes only JS. We have an HTML doc but our XML files aren't really close to that.

@oriadam I think all XML files should be aligned with each other so they should all use 2-space indentation.

entries/jQuery.getScript.xml
```diff
@@ -1,128 +1,136 @@
...
+$( "div.log" )
+	.ajaxError( function( e, jqxhr, settings, exception ) {
+		if ( settings.dataType === "script" ) {
+			$( this ).text( "Triggered ajaxError handler." );
+		}
+	} );
+		</code></pre>
+		<h4 id="caching-requests">Caching Responses</h4>
+		<p>By default, <code>$.getScript()</code> sets the cache setting to <code>false</code>. This appends a timestamped query parameter to the request URL to ensure that the browser downloads the script each time it is requested. 
+			You can override this feature by using a settings object, as of jQuery 1.12:</p>
```
mgol: Why the white space change here? It's all in one paragraph anyway so there will be no space in the rendered text.

entries/jQuery.getScript.xml
```diff
@@ -1,128 +1,136 @@
...
-      .animate({
-        backgroundColor: "olive"
-      }, 1000 )
-      .delay( 500 )
-      .animate({
-        backgroundColor: "#00f"
-      }, 1000 );
-  });
+	// Override toString for backward compatibility with jQuery prior to 1.12 (though it will prevent browser caching)
+	settings.toString = function() {
```
mgol: This seems pretty low-level, only plugin writes should be concerned about maintaining such support. I'd cut this mention.

oriadam: I don't think so, I actually added it BECAUSE I was searching for this for regular use.
Unless I completely misunderstood your intention

dmethvin: I think the issue is that if you're using 1.x *at all* at this point, you should be using 1.12.  We have previously (years ago) discussed trying to document the differences in all versions and decided it was too confusing. As far as software projects go I don't think our docs are significantly different from other projects in that regard.

PR #1065, opened 2017-11-06
contextmenu: Menu button trigger not just on root

Herst:
See also <https://developer.mozilla.org/en-US/docs/Web/Events/contextmenu>.

On a side note, can the root element always be assumed to be `html`?

### DIFF
--- a/entries/contextmenu.xml
+++ b/entries/contextmenu.xml
@@ -24,7 +24,7 @@
   </signature>
   <longdesc>
     <p>This method is a shortcut for <code>.on( "contextmenu", handler )</code> in the first two variations, and <code>.trigger( "contextmenu" )</code> in the third.
-    The <code>contextmenu</code> event is sent to an element when the right button of the mouse is clicked on it, but before the context menu is displayed. In case the context menu key is pressed, the event is triggered on the <code>html</code> element. Any HTML element can receive this event.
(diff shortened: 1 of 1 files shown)


AurelioDeRosa:
Thank you for contributing. In an HTML document, the root will always be the html element.

mgol:
@AurelioDeRosa The PR's intention wasn't to specify if the `html` element is equivalent to the root element but to specify the keyboard-activated context menu fires the `contextmenu` event at the currently focused element. I'll reopen it then.

@Herst Have you verified that this behavior is true? I don't have a Windows laptop to check that and Macs don't have the context menu key.

AurelioDeRosa:
@mgol I know the PR wasn't about that, I was replied to the question asked in this very PR.

> On a side note, can the root element always be assumed to be html?

Also, I'm a Windows user and I did verify that this is true.

mgol:
Sorry, my bad. Somehow I thought you rejected it, not landed. Closing again then!

Herst:
@mgol Yes, I have confirmed the behavior and not just on Windows BTW.

Cocerning the root document question I asked because when reading the documentation here I remembered that are reset stylesheets which for some reason use `:root` instead of the `html` selector in some places.

Again on a side note, I am using the `contextmenu` event to reliably prevent long-press touches from causing the context menu to pop up on certain interactive elements. Sadly there seems to be no way to prevent specifically prevent the event calls which where the result of a long-tap action as opposed to a right-click, see: https://stackoverflow.com/q/47202631/2261442 I would be glad if someone here knows the right commitees or such where such feature requests could be brought up.

Issue #1066, opened 2017-11-11
callbacks.fireWith has no types defined for its arguments


dratini0:
I have been processing the documentation with a python script (thank you for making it machine readable, btw!), when I noticed that the function [`callbacks.fireWith`](https://api.jquery.com/callbacks.fireWith/) has no types defined for its arguments. Since this is the only function like this, I am assuming this is a mistake. By looking at it, I am assuming they should be `Anything`, but I am not confident, since I have never used the `callbacks` object.

gibson042:
Thanks for finding this! Per https://github.com/jquery/jquery/pull/3844 , we'd like to describe the code accurately by documenting `type="Anything"` for `context` and `type="ArrayLikeObject"` for `args` (and updating the text to match).

dratini0:
That is almost exactly what I did in #1067, but I think you are right here. I didn't pick up on the ArrayLikeObject part, though. I will update #1067 accordingly

PR #1067, opened 2017-11-11
Fix signatures of multiple functions

dratini0:
Following #1066, I have found several functions without a return type. I have assigned appropriate types to these. I have additionally figured out the signature of callbacks.callWith, and fixed that as well. The documentation had a mistake in it, it stated that if the argument is not an array, the callbacks will receive a single argument, but this is not true, it is in fact passed as the second argument of apply.

Finally, I have not assigned a return type to jQuery.error, as it never returns therefore has no return type.

I agree to the CLA. Do I need any paperwork? EDIT: Figured it out.

### DIFF
--- a/entries/callbacks.fireWith.xml
+++ b/entries/callbacks.fireWith.xml
@@ -3,11 +3,13 @@
   <title>callbacks.fireWith()</title>
   <signature>
     <added>1.7</added>
-    <argument name="context" optional="true" type="Anything">
+    <argument name="context" optional="true" type="Object">
       <desc>A reference to the context in which the callbacks in the list should be fired.</desc>
     </argument>
-    <argument name="args" optional="true" type="Array">
-      <desc>An array of arguments to pass to the callbacks in the list. If omitted or undefined, no arguments will be passed.</desc>
+    <argument name="args" optional="true">
+      <desc>An argument, or array of arguments, to pass to the callbacks in the list.</desc>
+      <type name="Array"/>
+      <type name="Anything"/>
     </argument>
   </signature>
   <desc>Call all callbacks in a list with the given context and arguments.</desc>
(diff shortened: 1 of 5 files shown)


entries/callbacks.fireWith.xml
```diff
@@ -3,11 +3,11 @@
...
   <signature>
     <added>1.7</added>
-    <argument name="context" optional="true">
+    <argument name="context" optional="true" type="Anything">
       <desc>A reference to the context in which the callbacks in the list should be fired.</desc>
     </argument>
-    <argument name="args" optional="true">
-      <desc>An argument, or array of arguments, to pass to the callbacks in the list.</desc>
+    <argument name="args" optional="true" type="Array">
+      <desc>An array of arguments to pass to the callbacks in the list. If omitted or undefined, no arguments will be passed.</desc>
```
kswedberg: Hey @dratini0 , thanks for the PR! I don't think this change is correct, though. Looks to me like `args` can be "Anything" — an argument or array of arguments like the doc originally stated.

```
args = args || [];
args = [ context, args.slice ? args.slice() : args ];
queue.push( args );
```
https://github.com/jquery/jquery/blob/3e902a812014e8a6980e08599c4840b1cb969f7c/src/callbacks.js#L208-L210

dratini0: Take a look at the output of this fiddle then: https://jsfiddle.net/48Lmdmmf/

I have missed line 208, and assumed this was the intended behaviour. However, now I see that we are looking at a bug in jQuery, not in the documentation. I will put a pull request through shortly for this:

```js
args = [ context, args.slice ? args.slice() : [args] ];
```

I will also update this PR accordingly.

gibson042:
LGTM. @kswedberg?

kswedberg:
LGTM, too, @gibson042 . You wanna merge the PR? I would do it, but I can't seem to successfully run `grunt deploy` with my local setup anymore.

gibson042:
Same, but I'll give it a shot and :crossed_fingers:

kswedberg:
thank you, @gibson042 !

PR #1069, opened 2017-12-06
jQuery.cssHooks: remove the "O" Opera prefix

mgol:
jQuery 3.0+ doesn't support Opera Presto (versions <=12.x) so including its
prefix in the docs doesn't make much sense.


AurelioDeRosa:
LGTM

PR #1071, opened 2017-12-11
Update description to include return value information

mewelling:
The `.unwrap()` method returns whatever content was unwrapped. Adding this to the description so that the behavior is documented.


AurelioDeRosa:
Thank you very much for contributing to our project! I've merged your PR.

PR #1088, opened 2018-03-23
Make a minor correction

GeneralFailer:
I think the capitalization made dataType being a property name a bit unclear.


mgol:
Landed, thanks!

PR #1105, opened 2018-09-01
Fix #1022

### LINKED ISSUE #1022: [AJAX] add note about redirect failure due to differing origins
Something like "a 3xx redirect status will only be seen if the redirect is to another domain."

Migrated from https://github.com/jquery/jquery/issues/3467

zyphlar:
Trying to fix https://github.com/jquery/api.jquery.com/issues/1022

### DIFF
--- a/entries/jQuery.ajax.xml
+++ b/entries/jQuery.ajax.xml
@@ -160,7 +160,7 @@ $.ajax({
   }
 });
         </code></pre>
-          <p>If the request is successful, the status code functions take the same parameters as the success callback; if it results in an error (including 3xx redirect), they take the same parameters as the <code>error</code> callback.</p>
+          <p>If the request is successful, the status code functions take the same parameters as the success callback; if it results in an error (including 3xx redirects to another domain), they take the same parameters as the <code>error</code> callback. (3xx redirects to the same domain are generally handled transparently.)</p>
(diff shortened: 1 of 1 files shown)


entries/jQuery.ajax.xml
```diff
@@ -160,7 +160,7 @@ $.ajax({
   }
 });
         </code></pre>
-          <p>If the request is successful, the status code functions take the same parameters as the success callback; if it results in an error (including 3xx redirect), they take the same parameters as the <code>error</code> callback.</p>
+          <p>If the request is successful, the status code functions take the same parameters as the success callback; if it results in an error (including 3xx redirects to another domain), they take the same parameters as the <code>error</code> callback. (3xx redirects to the same domain are generally handled transparently.)</p>
```
AurelioDeRosa: Can you be more explicit about what "generally" means here?

dmethvin: "generally handled transparently" is vague, I agree. Perhaps "3xx redirects to the same domain are handled by the browser and not contrlled by jQuery."

mgol:
@zyphlar Do you plan to finish the PR?

mgol (changes requested):
Marking Request changes to make it explicit this is not ready

Issue #1107, opened 2018-09-17
Flag 'includeMargin' not documented but actually exists for setters outerWidth(value) and outerHeight(value)


YuriyYev:
Tested on jQuery 3.2.1.
Calling $(element).outerWidth(value, true) actually accounts for margins on element (and doesn't account when second arguement is not true). This fact (presence of flag) is not present in documentation at https://api.jquery.com/outerWidth/  for ".outerWidth( value )"

AurelioDeRosa:
I think this was specified a long time ago:

 > A Boolean indicating whether to include the element's margin in the calculation.

dmethvin:
In the docs we have a signature for the setter `.outerWidth(number)` and the getter `.outerWidth(boolean)` but not the setter `.outerWidth(number, boolean)`. We do have [tests for that though](https://github.com/jquery/jquery/blob/cf84696fd1d7fe314a11492606529b5a658ee9e3/test/unit/dimensions.js#L460-L522). So I suppose we need to document that case? Be aware that determining the dimensions for that case may force layout but I guess we'll let people figure that out themselves.

AurelioDeRosa:
I'll submit a PR tomorrow to address this issue.

PR #1133, opened 2019-05-06
Document optional includeMargin argument

AurelioDeRosa:
Fixes gh-1107


AurelioDeRosa:
@mgol agree on the Boolean trap it was implemented. As you mentioned, we're just documenting what has been there since a few years now.

PR #1139, opened 2019-07-17
a tiny fix: {a,i}s passed as

ryenus:
From `as passed` to  <code>is passed</code> as in:

    When a string as passed as an argument


mgol:
Thanks!

ryenus:
Welcome! Glad to see it's already in effect, cheers :-)

Issue #1144, opened 2019-08-28
Document that positional selectors are deprecated since jQuery 3.4.0


mgol:
Also, stop using them in examples on other API pages.

PR #1145, opened 2019-08-29
All: Mark positional selectors as deprecated since jQuery 3.4.0

mgol:
Positional selectors: `:first`, `:last`, `:eq`, `:even`, `:odd`, `:lt`, `:gt`, and `:nth` are deprecated since jQuery 3.4.0. This is now reflected in the docs & most deprecated pseudo-classes are not used in examples on other API pages.

The only exception is `:even` & `:odd` which are still used on some pages. I wasn't sure whether I should migrate them to a custom implementation using `.filter()` or whether we want `.even()` & `.odd()` in Core. I did a quick implementation to check the size impact & that would add 25 bytes to the gzipped file.

Fixes gh-1144


mgol:
This is the implementation that adds +25 bytes: https://github.com/mgol/jquery/commit/f45409708c6f307bf923440d09f6f2fd86df9a9d
```js
	even: function() {
		return this.filter( function( i ) {
			return ( i + 1 ) % 2;
		} );
	},

	odd: function() {
		return this.filter( function( i ) {
			return i % 2;
		} );
	},
```

mgol:
OK, I'm going to merge it; we can discuss the `.even()`/`.odd()` issue separately.

mgol:
I extracted the `:even`/`:odd` issue to #1147.

PR #1146, opened 2019-08-31
Clarify that errorThrown and statusText will be empty strings in HTTP/2

### LINKED ISSUE #1063: HTTP/2 and errorThrown or statusText
Hello!

In HTTP/2 errorThrown and statusText will be the empty strings:

>  HTTP/2 does not define a way to carry the version or reason phrase that is included in an HTTP/1.1 status line.
https://tools.ietf.org/html/rfc7540#section-8.1.2.4

I think that it should be added to this page https://api.jquery.com/jQuery.ajax/.

vkrol:
Fixes #1063.

### DIFF
--- a/entries/jQuery.ajax.xml
+++ b/entries/jQuery.ajax.xml
@@ -109,7 +109,7 @@ $.ajax({
         <argument name="jqXHR" type="jqXHR"/>
         <argument name="textStatus" type="String"/>
         <argument name="errorThrown" type="String"/>
-        <desc>A function to be called if the request fails. The function receives three arguments: The jqXHR (in jQuery 1.4.x, XMLHttpRequest) object, a string describing the type of error that occurred and an optional exception object, if one occurred. Possible values for the second argument (besides <code>null</code>) are <code>"timeout"</code>, <code>"error"</code>, <code>"abort"</code>, and <code>"parsererror"</code>. When an HTTP error occurs, <code>errorThrown</code> receives the textual portion of the HTTP status, such as "Not Found" or "Internal Server Error."  <strong>As of jQuery 1.5</strong>, the <code>error</code> setting can accept an array of functions. Each function will be called in turn.  <strong>Note:</strong> <em>This handler is not called for cross-domain script and cross-domain JSONP requests.</em> This is an <a href="/Ajax_Events/">Ajax Event</a>.</desc>
+        <desc>A function to be called if the request fails. The function receives three arguments: The jqXHR (in jQuery 1.4.x, XMLHttpRequest) object, a string describing the type of error that occurred and an optional exception object, if one occurred. Possible values for the second argument (besides <code>null</code>) are <code>"timeout"</code>, <code>"error"</code>, <code>"abort"</code>, and <code>"parsererror"</code>. When an HTTP error occurs, <code>errorThrown</code> receives the textual portion of the HTTP status, such as "Not Found" or "Internal Server Error." (it may be an empty string in HTTP/2) <strong>As of jQuery 1.5</strong>, the <code>error</code> setting can accept an array of functions. Each function will be called in turn.  <strong>Note:</strong> <em>This handler is not called for cross-domain script and cross-domain JSONP requests.</em> This is an <a href="/Ajax_Events/">Ajax Event</a>.</desc>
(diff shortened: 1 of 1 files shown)


entries/jQuery.ajax.xml
```diff
@@ -109,7 +109,7 @@ $.ajax({
...
+        <desc>A function to be called if the request fails. The function receives three arguments: The jqXHR (in jQuery 1.4.x, XMLHttpRequest) object, a string describing the type of error that occurred and an optional exception object, if one occurred. Possible values for the second argument (besides <code>null</code>) are <code>"timeout"</code>, <code>"error"</code>, <code>"abort"</code>, and <code>"parsererror"</code>. When an HTTP error occurs, <code>errorThrown</code> receives the textual portion of the HTTP status, such as "Not Found" or "Internal Server Error." (it will be an empty string in HTTP/2) <strong>As of jQuery 1.5</strong>, the <code>error</code> setting can accept an array of functions. Each function will be called in turn.  <strong>Note:</strong> <em>This handler is not called for cross-domain script and cross-domain JSONP requests.</em> This is an <a href="/Ajax_Events/">Ajax Event</a>.</desc>
```
dmethvin: Do you have an authoritative source verifying that HTTP2 implementations on all browsers cannot return strings for this information? I don't have a problem saying that they "may be blank" for example, but this change implies that they must be blank and jQuery doesn't control this.

vkrol: > Do you have an authoritative source verifying that HTTP2 implementations on all browsers cannot return strings for this information?

No. For example, Firefox does not return an empty string in violation of the specification. There is the open bug https://bugzilla.mozilla.org/show_bug.cgi?id=1397646. So I will replace it with "may be blank" as you suggested?

dmethvin: The thing we want to guard against is saying that it *will* be something when we simply pass through whatever the browser provides. how about adding `"(which may be an empty string)"` here? What was the case that led you to need this clarification?

vkrol: > how about adding "(which may be an empty string)" here?

I agree with that.

> What was the case that led you to need this clarification?

We encountered a couple of production errors after we updated from HTTP/1 to HTTP/2 because we had comparisons with `statusText` and `errorThrown` in our code.

vkrol: I changed the wording to `it may be an empty string in HTTP/2`.

entries/jQuery.ajax.xml
```diff
@@ -109,7 +109,7 @@ $.ajax({
...
+        <desc>A function to be called if the request fails. The function receives three arguments: The jqXHR (in jQuery 1.4.x, XMLHttpRequest) object, a string describing the type of error that occurred and an optional exception object, if one occurred. Possible values for the second argument (besides <code>null</code>) are <code>"timeout"</code>, <code>"error"</code>, <code>"abort"</code>, and <code>"parsererror"</code>. When an HTTP error occurs, <code>errorThrown</code> receives the textual portion of the HTTP status, such as "Not Found" or "Internal Server Error." (it may be an empty string in HTTP/2) <strong>As of jQuery 1.5</strong>, the <code>error</code> setting can accept an array of functions. Each function will be called in turn.  <strong>Note:</strong> <em>This handler is not called for cross-domain script and cross-domain JSONP requests.</em> This is an <a href="/Ajax_Events/">Ajax Event</a>.</desc>
```
mgol: I'd change:

> (it may be an empty string in HTTP/2)

to:

> (in HTTP/2 it may instead be an empty string)

entries/jQuery.ajax.xml
```diff
@@ -290,7 +290,7 @@ jqxhr.always(function() {
         <code>status</code>
       </li>
       <li>
-        <code>statusText</code>
+        <code>statusText</code> (it may be an empty string in HTTP/2)
```
mgol:
```suggestion
        <code>statusText</code> (may be an empty string in HTTP/2)
```

dmethvin:
Changes LGTM

vkrol:
@mgol thanks! Done.

mgol:
@vkrol Thanks!

Issue #982, opened 2016-09-19
Upgrade examples to jQuery 3


mgol:
Currently our examples use the really, really old jQuery 1.10.2 (it's over 3 years old). We should update them to jQuery 3 (currently 3.1.0). This means we will have to remove examples using removed APIs but I think we should do that anyway; such entries should be short and mostly focus on what to switch to, we shouldn't making it easier for people to use deprecated APIs.

The jQuery used in examples is included here: https://github.com/jquery/api.jquery.com/blob/a6f6b37/entries2html.xsl#L15.

mgol:
Note that the jQuery used by examples is a different one to the one used by our sites; the latter is managed globally in the `jquery-wp-content` repo, here: https://github.com/jquery/api.jquery.com/blob/a6f6b37/entries2html.xsl#L15

mgol:
The issue about upgrading jQuery on our sites is here: https://github.com/jquery/jquery-wp-content/issues/400

cc @dmethvin @timmywil

mgol:
PR: #1148

PR #1148, opened 2019-09-02
All: Upgrade jQuery used in examples from 1.10.2 to 3.4.1, remove deprecated examples

mgol:
The upgrade is necessary to be able to write examples for newer APIs.
Examples for removed APIs have been removed as they'd no longer work.

Fixes gh-982


mgol:
jQuery 1.10.2 was released roughly 6 years before 3.4.1! That's some upgrade delay... 😅

PR #1150, opened 2019-09-30
jQuery.getJSON: Change the removed jqXHR.complete to jqXHR.always

AndySky21:
Changed the "second complete" method according to Promise interface implementation


mgol:
Thanks!

Issue #1153, opened 2019-11-13
jQuery.removeData and jQuery.queue return types


edg2s:
In the documentation for the **utilities** [jQuery.removeData](https://api.jquery.com/jQuery.removeData/) and [jQuery.queue](https://api.jquery.com/jQuery.queue/) it is suggested that the return type may be a jQuery collection.

Looking at the code, this is clearly not the case.

jQuery.removeData never returns anything:
```js
removeData: function( elem, name ) {
	dataUser.remove( elem, name );
},
```
jQuery.removeData always returns the queue (an Array) or nothing (if `elem` is falsey):
```js
queue: function( elem, type, data ) {
	var queue;

	if ( elem ) {
		type = ( type || "fx" ) + "queue";
		queue = dataPriv.get( elem, type );

		// Speed up dequeue by getting out quickly if this is just a lookup
		if ( data ) {
			if ( !queue || Array.isArray( data ) ) {
				queue = dataPriv.access( elem, type, jQuery.makeArray( data ) );
			} else {
				queue.push( data );
			}
		}
		return queue || [];
	}
},
```

mgol:
Thanks for the issue.

> jQuery.removeData always returns the queue

I think you meant `jQuery.queue` here?

mgol:
Fixed by #1154 (81f438d5dfc02fcdbfed08f3caa177c23404244a) & #1155 (b1951ddfd19ceae1746bb91e1f8e5de6486ddfda).

PR #1154, opened 2019-11-13
jQuery.removeData always returns undefined

edg2s:
Not to be confused with the removeData method, which does return a jQuery collection.

See #1153


mgol:
Thanks!

PR #1155, opened 2019-11-13
jQuery.queue always returns an array

edg2s:
See #1153


mgol:
Thanks! BTW, please see the commit message format in the commit that was merged, I fixed it to adhere to our conventions.

PR #1162, opened 2020-05-05
jQuery.param: Document standaloneKeys argument

jayaddison:
This API documentation update accompanies a proposed change in https://github.com/jquery/jquery/pull/4708


mgol:
Thanks for the PR. However, per https://github.com/jquery/jquery/issues/4707#issuecomment-651217531 we decided not to make the change.

Issue #1164, opened 2020-07-13
jQuery.extend: separate one arg sig from 2+ args sig


timmywil:
See https://github.com/jquery/jquery/issues/4748

mgol:
PR: #1170

PR #1170, opened 2020-07-22
jQuery.extend: Separate the one-argument signature from the multiple-arguments one

mgol:
This is done to reduce the confusion as the single-argument signature behaves
completely differently than the other one.

Ref jquery/jquery#4748
Fixes gh-1164

### DIFF
--- a/entries/jQuery.extend.xml
+++ b/entries/jQuery.extend.xml
@@ -4,9 +4,9 @@
   <signature>
     <added>1.0</added>
     <argument name="target" type="Object">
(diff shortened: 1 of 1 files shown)


mgol:
Before:
Screen Shot 2020-07-22 at 13 57 20

After:
Screen Shot 2020-07-22 at 13 57 32

mgol:
Is the order fine or would it be better to have the separated single-arg signature immediately follow the first one?

mgol:
> @mgol fwiw, I like the way you have it, because the first 2 now seem more closely related conceptually.

Yeah, that was my thinking as well. I'm glad you like it. :)

PR #1175, opened 2020-08-20
jQuery.trim: Describe differences between String#trim & jQuery.trim

### LINKED ISSUE #1174: String.prototype.trim() isn't 100% compatible with jQuery.trim()
`jQuery.trim(null)` returns an empty string while `null.trim()` throws the error "Uncaught TypeError: Cannot read property 'trim' of null" 

The documentation page implies that it is a direct replacement. One should be aware of the consequences. It might help to note the potential issues when removing calls to $.trim() for all the users migrating to jQuery 3.5+.

`String.prototype.trim.call(null)` also throws "Uncaught TypeError: String.prototype.trim called on null or undefined".

ssg:
This fixes #1174

### DIFF
--- a/entries/jQuery.trim.xml
+++ b/entries/jQuery.trim.xml
@@ -10,7 +10,10 @@
   <desc>Remove the whitespace from the beginning and end of a string.</desc>
   <longdesc>
     <div class="warning">
-      <p>Note: This API has been deprecated in jQuery 3.5; please use the native <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim"><code>String.prototype.trim</code></a> method instead.</p>
+      <p>Note: This API has been deprecated in jQuery 3.5; please use the native <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim"><code>String.prototype.trim</code></a> method instead.
+      Unlike <code>jQuery.trim</code>, <code>String.prototype.trim</code> does not work with types other than strings (<code>null</code>, <code>undefined</code>, <code>Number</code>). 
+      Make sure that your code is compatible when migrating.
+      </p>
     </div>
     <p>The <code>$.trim()</code> function removes all newlines, spaces (including non-breaking spaces), and tabs from the beginning and end of the supplied string. If these whitespace characters occur in the middle of the string, they are preserved.</p>
   </longdesc>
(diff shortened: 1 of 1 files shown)


entries/jQuery.trim.xml
```diff
@@ -10,7 +10,10 @@
   <desc>Remove the whitespace from the beginning and end of a string.</desc>
   <longdesc>
     <div class="warning">
-      <p>Note: This API has been deprecated in jQuery 3.5; please use the native <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim"><code>String.prototype.trim</code></a> method instead.</p>
+      <p>Note: This API has been deprecated in jQuery 3.5; please use the native <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim"><code>String.prototype.trim</code></a> method instead.
+      Unlike <code>jQuery.trim</code>, <code>String.prototype.trim</code> does not work with types other than strings (<code>null</code>, <code>undefined</code>, <code>Number</code>). 
+      Make sure that your code is compatible when migrating.
+      </p>
```
mgol, line 16: Do you want that to be two paragraphs? It makes sense but then you'll need to insert those markers

mgol, line 16: Please keep the closing tag on the same line as end of its contents.

timmywil:
I like it.

mgol (commented):
Thanks for the PR! I added a few suggestions

entries/jQuery.trim.xml
ssg: No, I assumed it to be a single paragraph.

ssg: Done!

entries/jQuery.trim.xml
```diff
@@ -10,7 +10,9 @@
   <desc>Remove the whitespace from the beginning and end of a string.</desc>
   <longdesc>
     <div class="warning">
-      <p>Note: This API has been deprecated in jQuery 3.5; please use the native <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim"><code>String.prototype.trim</code></a> method instead.</p>
+      <p>Note: This API has been deprecated in jQuery 3.5; please use the native <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim"><code>String.prototype.trim</code></a> method instead.
+      Unlike <code>jQuery.trim</code>, <code>String.prototype.trim</code> does not work with types other than strings (<code>null</code>, <code>undefined</code>, <code>Number</code>).
```
mgol, line 14: If this is a single paragraph, let's drop the new line here & write it all in one line.

ssg: Done

mgol (approved):
Looks good, thanks!

PR #1183, opened 2021-04-02
addClass:removeClass:toggleClass: Update function signatures

kidonng:
[From jQuery 3.3.0](https://blog.jquery.com/2018/01/19/jquery-3-3-0-a-fragrant-bouquet-of-deprecations-and-is-that-a-new-feature/), if using a function as a argument for `.addClass`, `.removeClass` and `.toggleClass`, the function is allowed to return a `Array`. However, this is not reflected in the return types.

### DIFF
--- a/entries/addClass.xml
+++ b/entries/addClass.xml
@@ -25,7 +25,7 @@
   <signature>
     <added>3.3</added>
     <argument name="function" type="Function">
-      <desc>A function returning one or more space-separated classes or an array of classes to be added to the existing class name(s). Receives the index position of the element in the set and the existing class name(s) as arguments. Within the function, <code>this</code> refers to the current element in the set.</desc>
+      <desc>A function returning one or more space-separated class names or an array of class names to be added to the existing class name(s). Receives the index position of the element in the set and the existing class name(s) as arguments. Within the function, <code>this</code> refers to the current element in the set.</desc>
       <argument name="index" type="Integer" />
       <argument name="currentClassName" type="String" />
       <return type="String"/>
--- a/entries/addClass.xml
+++ b/entries/addClass.xml
@@ -16,6 +16,15 @@
   <signature>
     <added>1.4</added>
     <argument name="function" type="Function">
+      <desc>A function returning one or more space-separated class names to be added to the existing class name(s). Receives the index position of the element in the set and the existing class name(s) as arguments. Within the function, <code>this</code> refers to the current element in the set.</desc>
+      <argument name="index" type="Integer" />
+      <argument name="currentClassName" type="String" />
+      <return type="String"/>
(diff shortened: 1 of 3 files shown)


mgol (changes requested):
Thanks for the PR.

This signature was added in jQuery 3.3, ass you've notice, so it cannot be added to the entry for the 1.4 release. You'll need to create a new one.

kidonng:
Did you mean I need to add a new signature like this?

```xml
  <signature>
    <added>3.3</added>
    <argument name="function" type="Function">
      <desc>A function returning one or more space-separated classes or an array of classes to be added to the existing class name(s). Receives the index position of the element in the set and the existing class name(s) as arguments. Within the function, <code>this</code> refers to the current element in the set.</desc>
      <argument name="index" type="Integer" />
      <argument name="currentClassName" type="String" />
      <return type="String"/>
      <return type="Array"/>
    </argument>
  </signature>
```

But I noticed the new signature is already mentioned in the description:

> A function returning one or more space-separated classes **or an array of classes** to be added to the existing class name(s).

mgol:
> Did you mean I need to add a new signature like this? (...)

Yes; this will be some duplication but it is a new signature in the end so that seems to be the best way.

> But I noticed the new signature is already mentioned in the description:
> 
> > A function returning one or more space-separated classes **or an array of classes** to be added to the existing class name(s).

Nice finding! This is a mistake I made in https://github.com/jquery/api.jquery.com/commit/13b661bd061b7c341cc74a44e6653937ebb9f556#diff-b61e9e36c5e0af399886d0e394d36f012a444c831eb2331ae67118dcf437b813; that part of the note should be removed from the `1.4` entry and only be present in the `3.3` one. Can you do such modifications? Other files have similar issues, I think.

mgol (commented):
Just minor remarks: if we use "class names" in descriptions, let's use it consistently across both 1.4 & 3.3. descriptions instead of "classes".

Also, for consistency I'd change the `toggleClass` descriptions from "A function that returns" to "A function returning", or vice versa.

kidonng:
> Just minor remarks: if we use "class names" in descriptions, let's use it consistently across both 1.4 & 3.3. descriptions instead of "classes".

I only changed the ones covered in the related sections, but there are more instances of `classes`, should I change them as well?

mgol:
@kidonng 

> I only changed the ones in the related sections, but there are more instances of "classes", should I change them as well?

That's a bit far from the original change here so we can leave it for potential future improvements in separate PRs. I don't have a definite opinion on what's better: `class names` is more explicit but `classes` is shorter and a class is usually treated as a string by which it's identified.

mgol (approved):
Thank you, LGTM!

Issue #1184, opened 2021-05-01
.addClass section shows unexpected error message


anderlaw:
When checking out the jQuery's API document,  I found a very  confusing message, which is marked in the image below:

![Snipaste_2021-05-01_14-44-03](https://user-images.githubusercontent.com/21139011/116773963-b5128d00-aa8b-11eb-98be-c45f09a7499f.png)

I try to figure out why based on the Error Message:`ERROR: A single return element is expected`, then I found there are two return types `String` or `Array`:
```
 <signature>
    <added>3.3</added>
    <argument name="function" type="Function">
      <desc>A function returning one or more space-separated class names or an array of class names to be added to the existing class name(s). Receives the index position of the element in the set and the existing class name(s) as arguments. Within the function, <code>this</code> refers to the current element in the set.</desc>
      <argument name="index" type="Integer" />
      <argument name="currentClassName" type="String" />
      <return type="String"/>
      <return type="Array"/>
    </argument>
  </signature>
```
then I try to find out why it didn't support two return types, because it should be, finally located to an npm package named `grunt-jquery-content` 's `entries2html-base.xsl` file：
and when I modified the code from
```
	<!-- a return element -->
	<xsl:if test="return">
		<xsl:for-each select="return">
			<xsl:if test="position() &gt; 1">
				<strong>ERROR: A single return element is expected</strong>
			</xsl:if>
			<xsl:call-template name="render-types"/>
		</xsl:for-each>
	</xsl:if>
```
to
```
	<!-- a return element -->
	<xsl:if test="return">
		<xsl:for-each select="return">
			<xsl:if test="position() &gt; 1">
				<span> | </span>
			</xsl:if>
			<xsl:call-template name="render-types"/>
		</xsl:for-each>
	</xsl:if>
```
the error message missed, and everything includes links works fine:
![fine](https://user-images.githubusercontent.com/21139011/116774531-c067b780-aa8f-11eb-8e06-9bae6f185c0a.png)

mgol:
Nice catch! Interesting, top-level returns work which you can see e.g. on the [`.val()`](https://api.jquery.com/val/) page, it's just returns in functions that are broken.

Would you like to submit a PR against `grunt-jquery-content`? You can ping me for a review.

anderlaw:
thanks for your response, I just made a PR [here](https://github.com/jquery/grunt-jquery-content/pull/77#partial-pull-merging), thanks

mgol:
[...]

PR #1187, opened 2021-05-13
removeClass: Separate the signature with no parameters

mgol:
That signature removes all classes which wasn't that clear from the original
signatures until one read the description.

I recommend reviewing with whitespace changes disabled as the whole file had to be indented.

This has been a point of confusion at my company recently and even I didn't quickly find that it was actually documented as the `removeClass( [ className ] )` signature was describing two essentially different signatures.

### DIFF
--- a/entries/removeClass.xml
+++ b/entries/removeClass.xml
@@ -36,7 +36,6 @@
     </signature>
     <desc>Remove a single class or multiple classes from each element in the set of matched elements.</desc>
     <longdesc>
-      <p>If a class name is included as a parameter, then only that class will be removed from the set of matched elements. If no classes are specified in the parameter, all classes will be removed.</p>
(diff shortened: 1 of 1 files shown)


timmywil (approved):
LGTM. I reviewed with whitespace off. Side note: I wish I could review this rendered as it seems easier to miss things in diffs.

mgol:
@timmywil I pushed one more commit to remove a redundant sentence.

As for how it looks after the changes, this is the top of the page:
Screen Shot 2021-05-17 at 18 30 40
and this is where the second, new entry starts:
Screen Shot 2021-05-17 at 18 30 58

You can compare that to what we have now. Please have a look.

timmywil:
This looks familiar and may be fixed elsewhere, but there's an issue with the return on `removeClass(function)`.
image

mgol:
@timmywil Yeah, this is tracked in #1184, there's a PR I need to review.

PR #1188, opened 2021-05-13
CSS: Make it clear !important is not guaranteed to not work with `.css()`

mgol:
This is so that we can add support in a minor release without treating it as a breaking change.


timmywil (approved):
LGTM

PR #1189, opened 2021-05-13
removeProp and prop: clarify recommendations regarding native properties

timmywil:
Ref https://github.com/jquery/jquery/issues/4887


mgol (approved):

akshaygao (commented):
Similar such documentation is available in .prop().  The highlighted text lead me to the confusion.

Important: the .removeProp() method should not be used to set these properties to false. **Once a native property is removed, it cannot be added again**. See .removeProp() for more information.

mgol (approved):

Issue #1207, opened 2022-05-22
$.ajax() documentation should be fixed in a case where script isn't necessarily executed yet


vlakoff:
[`$.getScript()`](https://api.jquery.com/jquery.getscript/) returns a jqXHR object that may be used as so:

```javascript
$.getScript( '/foobar.js' ).done( function ( script, textStatus ) {
    // script has been loaded and executed
} );
```

However, it can also be provided a second, optional `success` argument:

```javascript
$.getScript( '/foobar.js', function ( data, textStatus, jqXHR ) {
    // script has been loaded, but not necessarily executed yet
} );
```

Notice in the second snippet that the script isn't necessarily executed yet.

<br> This subtle, but important, difference is noted in the [`$.getScript()` documentation](https://api.jquery.com/jquery.getscript/#success-callback):

> The callback is fired once the script has been loaded but not necessarily executed.

However, `$.getScript()` is just a shorthand method that calls [`$.ajax()`](https://api.jquery.com/jquery.ajax/), and the latter seems to have a mistake in [its documentation](https://api.jquery.com/jquery.ajax/#data-types):

> If `script` is specified, `$.ajax()` will execute the JavaScript that is received from the server before passing it on to the success handler as a string.

<br> To summarize, if I correctly figured out the behavior of `$.ajax()`, the remote script has been executed when using the returned jqXHR object, but not necessarily when using the callbacks (`success`, `complete`, etc.) passed to the `settings` argument.

Thus, the above line from the documentation should be fixed. You may also want to emphasize these two different behaviors.

mgol:
PRs to update the docs are welcome!

vlakoff:
Nowadays, with the second syntax from OP, internally the `success` argument is attached as a `done()` handler to the `jqHXR` instance; see [ajax.js#L670](https://github.com/jquery/jquery/blob/5d5ea015114092c157311c4948f7cc3d8c8e7f8a/src/ajax.js#L670).

That's the same `jqHXR` instance that gets returned and attached a `done()` handler with the first syntax from OP.

So, both syntaxes are equivalent.

Though, we should investigate if it has always been the case. Maybe the jQuery code has been modified at some point.

vlakoff:
I made some digging, and the `complete` argument gets attached to the returned jqXHR instance since these commits from late 2010 / early 2011:

* https://github.com/jquery/jquery/commit/ab3ba4a81252c4357a7aab5f24d765d41d47986e
[...]

PR #1209, opened 2022-10-19
jQuery.getScript: Script has been executed before the `success` callback

vlakoff:
When executing a callback provided to the optional `success` parameter of [`$.getScript()`](https://api.jquery.com/jquery.getscript/),
we can assume the script has been executed already.

Just like when using `.done()` on the returned jqXHR object.

See issue #1207.


mgol (approved):
Thank you!

PR #1214, opened 2022-12-19
jQuery.map:jQuery.uniqueSort: Accept array-like input, fix typos

mgol:
_1\. `jQuery.map`: Accept array-like input_

_2\. `jQuery.uniqueSort`: Accept array-like input, fix typos_

Apart from array-like inputs being officially allowed in
`jQuery.uniqueSort` now, in a few places in examples
it used to be referred to as `unique` instead of `uniqueSort`.

----- 

The second one is particularly interesting. For some reason, there's even an example directly claiming `jQuery.uniqueSort` doesn't work on array-likes when its code says something completely opposite - especially that it's sometimes used on jQuery collections. [We even test it on an "Arrayish" class](https://github.com/jquery/jquery/blob/024d87195ac46690023e2b0b308d4406a8a5a27e/test/unit/selector.js#L1957-L1962)!

### DIFF
--- a/entries/jQuery.map.xml
+++ b/entries/jQuery.map.xml
@@ -28,7 +28,7 @@
   <desc>Translate all items in an array or object to new array of items.</desc>
   <longdesc>
     <p>If you wish to process a jQuery object — for example, <code>$('div').map( callback );</code> — use <a href="/map/">.map()</a> instead. </p>
-    <p>The <code>$.map()</code> method applies a function to each item in an array or object and maps the results into a new array. <strong>Prior to jQuery 1.6</strong>, <code>$.map()</code> supports traversing <em>arrays only</em>. <strong>As of jQuery 1.6</strong> it also traverses objects.</p>
+    <p>The <code>$.map()</code> method applies a function to each item in an array or object and maps the results into a new array. <strong>Prior to jQuery 1.6</strong>, <code>$.map()</code> supports traversing <em>arrays and array-like objects only</em>. <strong>As of jQuery 1.6</strong> it also traverses objects.</p>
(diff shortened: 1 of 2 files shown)


entries/jQuery.map.xml
```diff
@@ -3,8 +3,8 @@
   <title>jQuery.map()</title>
   <signature>
     <added>1.0</added>
-    <argument name="array" type="Array">
-      <desc>The Array to translate.</desc>
+    <argument name="array" type="ArrayLikeObject">
+      <desc>The Array or an Array-like object to translate.</desc>
```
timmywil, line 7: Are we sure this is true for all versions back to 1.0?

mgol: @timmywil I've just checked. In versions `1.0`, `1.1`, `1.2`, `1.3`, `1.4` & `1.5` `jQuery.map` is written using a for loop that would work with array-likes as well. Moreover, jQuery 1.2 implemented `jQuery.fn.map` which uses `jQuery.map` on jQuery objects which are not arrays.

jQuery `1.6` added support for iterating on objects in `jQuery.map` so it had to detect "arrays" separately. The `1.6` detection code is a bit cryptic:
```js
// jquery objects are treated as arrays
isArray = elems instanceof jQuery || length !== undefined && typeof length === "number" && ( ( length > 0 && elems[ 0 ] && elems[ length -1 ] ) || length === 0 || jQuery.isArray( elems ) )
```
but it does detect array-likes. `1.9` extracted that logic and uses:
```js
isArray = isArraylike( elems ),
```
so even the util name has "array-like" in the name.

I think we can assume that array-likes have been supported since `1.0` even if that was not the original intent.

PR #1218, opened 2023-01-11
Update jQuery.ajax.xml: change option `type` to `method`

dom-rau:


entries/jQuery.ajax.xml
```diff
@@ -340,7 +340,7 @@ jqxhr.always(function() {
...
     <p>For more information on JSONP, see the <a href="https://bob.pythonmac.org/archives/2005/12/05/remote-json-jsonp/">original post detailing its use</a>.</p>
     <h4 id="sending-data-to-server">Sending Data to the Server</h4>
-    <p>By default, Ajax requests are sent using the GET HTTP method. If the POST method is required, the method can be specified by setting a value for the <code>type</code> option. This option affects how the contents of the <code>data</code> option are sent to the server. POST data will always be transmitted to the server using UTF-8 charset, per the W3C XMLHTTPRequest standard.</p>
+    <p>By default, Ajax requests are sent using the GET HTTP method. If the POST method is required, the method can be specified by setting a value for the <code>method</code> option. If you're using versions of jQuery prior to 1.9.0 you should use `type` option. This option affects how the contents of the <code>data</code> option are sent to the server. POST data will always be transmitted to the server using UTF-8 charset, per the W3C XMLHTTPRequest standard.</p>
```
mgol, line 343: We're not deprecating the `type` option so this wording is not ideal IMO. Can we just mention both and only indicate that the `method` one requires `1.9.0` or newer?

mgol (changes requested):
Thanks for the PR! I have one comment.

mgol:
@dominikarauk ping

PR #1225, opened 2023-04-27
All: Add missing jQuery versions to categories, update entries

mgol:
jQuery 1.10/2.0 & 1.11/2.1 now have dedicated version categories; the same applies to 3.6. A few other version descriptions have been tweaked.

The `.add( selection )` signature is now documented to have arrived in 1.1 instead of 1.3.2. Tests for this behavior got added in https://github.com/jquery/jquery/commit/a5f9108a2109b2ed5778af860b0928d8e6b0fdd2.

A cetegory in one of `before` & `after` signatures was updated from `1.10` to `1.10-and-2.0`, fixing a broken category link.

This should fix the `spider-check` that is broken right now on GitHub Actions: https://github.com/jquery/api.jquery.com/actions/runs/4789051019/jobs/8516452787

### DIFF
--- a/categories.xml
+++ b/categories.xml
@@ -439,11 +439,24 @@ var files = event.originalEvent.dataTransfer.files;
         <hr/>
       ]]></desc>
     </category>
+    <category name="Version 1.10 &amp; 2.0" slug="1.10-and-2.0">
+      <desc><![CDATA[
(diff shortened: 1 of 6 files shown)


mgol:
BTW, I didn't add a note about JSONP error responses now returning the JSON to the `jQuery.ajax` page as we highlighted in the [3.6.0 blog post](https://blog.jquery.com/2021/03/02/jquery-3-6-0-released/) as this was not actually a new feature - it worked before 3.5.0, broke in 3.5.0 and we brought it back in 3.6.0. I felt it's too niche to describe in detail peculiarities of such a temporary change on a page that is already large.

Krinkle (approved):

PR #1231, opened 2023-05-13
Build: Skip checking Twitter links in the hydra crawler

mgol:
Twitter pages now do 302-redirects to themselves for users without a specific cookie set which trips the crawler; avoid checking Twitter links by abusing the `exclude_scheme_prefixes` option of the crawler.

Since the project only accepts options in a form of a configuration file, we also need to clone the API repo to provide such a file.

### DIFF
--- a/.github/workflows/spider-check.yaml
+++ b/.github/workflows/spider-check.yaml
@@ -20,9 +20,18 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.repository_owner == 'jquery' }} # skip on forks
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout hydra-link-checker
+        uses: actions/checkout@v3
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/spider-check.yaml
```diff
@@ -20,9 +20,18 @@ jobs:
...
         with:
           repository: jquery/hydra-link-checker
           ref: v2.0.0
+          path: hydra
+
+      # Checkout the API repo as well to provide the config for hydra-link-checker
+      - name: Checkout API repo
+        uses: actions/checkout@v3
+        with:
+          path: api
```
mgol, line 34: Is there a better way to provide a single file to another step?

Krinkle: Not that I know of. LGTM.

Krinkle (approved):

PR #1237, opened 2023-07-24
Build: install libxml2 for lint task

timmywil:


timmywil:
Merging to fix the main branch. I shouldn't have committed the new workflow directly to the main branch from the GH actions GUI the first time, but this at least gets it passing.

Issue #1241, opened 2023-11-10
Mention of ‘timestamp’ in documentation for ‘cache’ parameter of ‘ajax’ is misleading


Rinzwind:
The documentation for the ‘cache’ parameter of ‘ajax’ says:

>  It works by appending "_={timestamp}" to the GET parameters.

This seems a bit misleading. I interpreted it as a timestamp of when the request is sent. But the value after `_=` is actually taken from a counter:

https://github.com/jquery/jquery/blob/f79d5f1a337528940ab7029d4f8bbba72326f269/dist/jquery.js#L9435

The counter is only initialized using `Date.now()`:

https://github.com/jquery/jquery/blob/f79d5f1a337528940ab7029d4f8bbba72326f269/dist/jquery.js#L8501

I got confused by that when analyzing nginx access logs and wondering why the ‘timestamps’ didn’t match up.

The documentation should probably just say that it works by appending "_={uid}".

mgol:
Thanks for the report. I transferred it to the API repo since it's a docs issue.

Would you like to submit a PR?

BTW, there are a few more similar entries to potentially replace on that same page.

Rinzwind:
Sure, I created: https://github.com/jquery/api.jquery.com/pull/1242

PR #1242, opened 2023-11-15
Change explanation of the ‘cache’ setting for ‘ajax’ and ‘getScript’ to say it appends a counter value rather than a timestamp

Rinzwind:
This pull request should fix issue #1241. It’s a straightforward replacement of ‘timestamp’ by ‘counter’. Saying ‘identifier’ seemed a bit vague. Though I guess ‘counter’ might also be misinterpreted as a count of the number of times the given URL has been requested.


mgol:
@Rinzwind please sign our CLA

PR #1248, opened 2024-07-12
Build: Refer to commits of GitHub Actions, upgrade them

mgol:
Also, set up automatic Dependabot updates of actions grouped into a single PR as Core does.

Ref jquery/jquery#5503

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -3,8 +3,9 @@
 name: Node.js CI
 
 on:
-  - push
-  - pull_request
+  pull_request:
+  push:
+    branches-ignore: "dependabot/**"
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


.github/workflows/node.js.yml
```diff
@@ -3,8 +3,10 @@
 name: Node.js CI
 
 on:
-  - push
-  - pull_request
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
```
mgol, line 9: For compat with jQuery, maybe it should be:
```suggestion
  pull_request:
  push:
    branches-ignore: "dependabot/**"
```
Thoughts?

mgol: Updated.

timmywil (approved):

PR #1251, opened 2024-09-29
Categories: add version support warnings to old versions

timmywil:
- also made formatting consistent for all deprecated/version categories

The styling for the warning messages is in the [companion PR in jquery-wp-content](https://github.com/jquery/jquery-wp-content/pull/468) (some of the styling already existed). I prefer to set these explicitly rather than try to generate them programmatically. And, this gives us a little side advantage of being able to modify the message for specific categories, if needed. I made use of that when there were multiple versions and pluralized the message.

Ref https://github.com/jquery/jquery-wp-content/pull/468
Ref https://github.com/jquery/jquery-wp-content/issues/462

### DIFF
--- a/categories.xml
+++ b/categories.xml
@@ -37,58 +37,75 @@
   <category name="Deprecated" slug="deprecated">
     <desc/>
     <category name="Deprecated 1.3" slug="deprecated-1.3">
-      <desc><![CDATA[All the aspects of the API that were deprecated in the corresponding version of jQuery.
+      <desc><![CDATA[
+        <div id="version-support-warning" class="warning"><i class="icon-info-sign"></i> <span>This version is no longer supported. Read more about our <a href="https://jquery.com/support/">Version Support</a>.</span></div>
+        <p>All the aspects of the API that were deprecated in the corresponding version of jQuery.</p>
         <p>For more information, see the <a href="https://blog.jquery.com/2009/01/14/jquery-1-3-released/">jQuery 1.3 Release Notes</a>.</p>
       ]]></desc>
     </category>
     <category name="Deprecated 1.4" slug="deprecated-1.4">
-      <desc><![CDATA[All the aspects of the API that were deprecated in the corresponding version of jQuery.
+      <desc><![CDATA[
+        <div id="version-support-warning" class="warning"><i class="icon-info-sign"></i> <span>This version is no longer supported. Read more about our <a href="https://jquery.com/support/">Version Support</a>.</span></div>
+        <p>All the aspects of the API that were deprecated in the corresponding version of jQuery.</p>
         <p>For more information, see the <a href="https://web.archive.org/web/20150119205819/http://jquery14.com/day-01/jquery-14">jQuery 1.4 Release Notes</a></p>
       ]]></desc>
     </category>
     <category name="Deprecated 1.7" slug="deprecated-1.7">
-      <desc><![CDATA[All the aspects of the API that were deprecated in the corresponding version of jQuery.
+      <desc><![CDATA[
+        <div id="version-support-warning" class="warning"><i class="icon-info-sign"></i> <span>This version is no longer supported. Read more about our <a href="https://jquery.com/support/">Version Support</a>.</span></div>
+        <p>All the aspects of the API that were deprecated in the corresponding version of jQuery.</p>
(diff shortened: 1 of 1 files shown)


categories.xml
```diff
@@ -37,58 +37,75 @@
...
+        <div id="version-support-warning" class="warning"><i class="icon-info-sign"></i> <span>These versions are no longer supported. Read more about our <a href="https://jquery.com/support/">Version Support</a>.</span></div>
+        <p>All the aspects of the API that were deprecated in the corresponding version of jQuery.</p>
         <p>For more information, see the Release Notes/Changelog at <a href="https://blog.jquery.com/2013/05/24/jquery-1-10-0-and-2-0-1-released/">https://blog.jquery.com/2013/05/24/jquery-1-10-0-and-2-0-1-released/</a></p>
       ]]></desc>
     </category>
     <category name="Deprecated 3.0" slug="deprecated-3.0">
-      <desc><![CDATA[All the aspects of the API that were deprecated in the corresponding version of jQuery.
+      <desc><![CDATA[
         <p>For more information, see the Release Notes/Changelog at <a href="https://blog.jquery.com/2016/06/09/jquery-3-0-final-released/">https://blog.jquery.com/2016/06/09/jquery-3-0-final-released/</a></p>
+        <p>All the aspects of the API that were deprecated in the corresponding version of jQuery.</p>
```
mgol, line 84: The above "For more information" sounds a bit weird when it's the first sentence.

This applies to many places.

categories.xml
```diff
@@ -326,193 +343,270 @@ var files = event.originalEvent.dataTransfer.files;
...
-        <p>For more information, see the Release Notes/Changelog at <a href="https://blog.jquery.com/2011/11/03/jquery-1-7-released/">https://blog.jquery.com/2011/11/03/jquery-1-7-released/</a></p>
+        <div id="version-support-warning" class="warning"><i class="icon-info-sign"></i> <span>This version is no longer supported. Read more about our <a href="https://jquery.com/support/">Version Support</a>.</span></div>
+        <p>All the aspects of the API that were added, or had a new signature added, in the corresponding version of jQuery.</p>
+        <p>jQuery 1.7.0 included:</p>
+        <ul>
+          <li>New event APIs: <code>.on()</code> and <code>.off()</code></li>
+          <li>Better Support for HTML5 in IE6/7/8</li>
+          <li><code>jQuery.Callbacks()</code></li>
+          <li>Toggling Animations Work Intuitively</li>
+        </ul>
```
mgol, line 499: I like a proper list here!

Krinkle:
I'm unsure about the banner on "version added" category pages. These predominantly contain pages about features that continue to exist in even the latest beta with no hint or sign that they're deprecated. Warning there seems deceptive and may confuse/break navigation flows and curiosity, specifically relating to how stable and mature something is based on when it was first added ("can I use").

+1 for adding to categories about deprecated and removed versions, which describes features in older versions indeed.

timmywil:
I considered that, but since you brought it up as well. Let me think on the wording and I'll make a proposal soon.

categories.xml
timmywil: Oops, that was unintentional.

timmywil:
I changed the version category pages to say this instead:

> Some APIs may have changed since this version was released. If using this version, find out about our <a href="https://jquery.com/support">version support</a>.

mgol:
> I changed the version category pages to say this instead:
> 
> > Some APIs may have changed since this version was released. If using this version, find out about our [version support](https://jquery.com/support).

It still doesn't fully feel right to me. The same sentence would make sense on version 3.7 as well once 4.0 gets released, even though we haven't removed support for this version. Unless you'd add such a banner on all 3.x versions as well as soon as 4.0.0 is out?

timmywil:
> Unless you'd add such a banner on all 3.x versions as well as soon as 4.0.0 is out?

That was the idea, yeah. When 4.0.0 is released, we'd add a similar notice to 3.x versions.

mgol:
> That was the idea, yeah. When 4.0.0 is released, we'd add a similar notice to 3.x versions.

OK, cool, that works for me then.

mgol (approved):

Krinkle (approved):

PR #1253, opened 2024-10-05
Build: update default deployment config

timmywil:


config-sample.json
```diff
@@ -1,5 +1,5 @@
 {
-  "url": "vagrant.api.jquery.com",
-  "username": "admin",
-  "password": "secret"
+  "url": "local.api.jquery.com",
```
mgol, line 2: We also need the scheme here, don't we?

mgol (changes requested):

config-sample.json
timmywil: It actually works either way for me, but I did think at one point that was better for Andre. I'll add it here.

Issue #1243, opened 2024-03-18
Document changes in boolean attributes treatment


mgol:
See https://github.com/jquery/jquery/pull/5452

gibson042:
The table in [`.attr( attributeName )`](https://api.jquery.com/attr/#attr1) will need a new row for 4.0+, and [`.attr( attributeName, value )`](https://api.jquery.com/attr/#attr2) will need updates to document the acceptability of a `false` value and its removal behavior which is similar to `null` behavior except with "aria-" attributes.

Proposed text for the latter:
> Note: because [ARIA](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA) attributes frequently associate behavior with "false" values that differs from attribute absence, passing `false` as the value for an attribute whose name starts with "aria-…" will stringify that value to "false" rather than remove the attribute. To guarantee removal of an attribute, provide `null` as the value or use the `.removeAttr()` method.

mgol:
PR: #1254

PR #1254, opened 2024-10-09
Attr: Document jQuery 4 boolean attribute changes

mgol:
Fixes gh-1243
Ref jquery/jquery#5452
Ref jquery/jquery-migrate#540

PR #1255 needs to land first.

~~I added a `4.0` category here as well. We are close to the release and I'd prefer start merging docs changes for jQuery 4.0 instead of waiting for the release. I added a banner to the category indicating it's a pre-release, we can update it if we release an RC and/or a stable release, but the docs would remain the same.~~

~~I'll base other docs PRs for 4.0 on this PR so that the category is there.~~


mgol:
OK, I extracted the category PR to #1255.

mgol:
Reopened at #1263.

Issue #1200, opened 2022-02-22
Document jQuery.fx.interval removal from jQuery 4.0.0


mgol:
It's been deprecated in jQuery 3.0.0.

Removal PR: https://github.com/jquery/jquery/pull/5017

mgol:
PR: #1256

Issue #1172, opened 2020-07-29
Document `toggleClass(boolean)` & `toggleClass(undefined)` are removed in 4.0


mgol:

mgol:
PR: #1256

PR #1256, opened 2024-10-09
All: Record APIs removed in 4.0

mgol:
Fixes gh-1172
Fixes gh-1200


mgol:
New version at #1262.

PR #1260, opened 2024-10-09
jQuery.cssNumber: Document the removal in 4.0

mgol:
This is a weird API as it was never deprecated since jQuery `3.x` uses it but jQuery `4.0.0` will not need it.


mgol:
css: Document no auto-appending `px` in jQuery 4.0

Also, document the removal of `jQuery.cssNumber` in jQuery 4.0.

mgol:
Reopened at #1261.

PR #1261, opened 2024-10-14
css: Document no auto-appending `px` in jQuery 4.0

mgol:
Also, document the removal of `jQuery.cssNumber` in jQuery 4.0.

`jQuery.cssNumber` is a weird API as it was never deprecated since jQuery `3.x` uses it but jQuery `4.0.0` will not need it.


mgol:
This is a new version of #1260.

entries/jQuery.cssNumber.xml
```diff
@@ -1,31 +1,16 @@
...
-      <li><code>opacity</code></li>
-      <li><code>zoom</code></li>
-      <li><code>lineHeight</code></li>
-      <li><code>widows</code> (added in jQuery 1.6)</li>
-      <li><code>orphans</code> (added in jQuery 1.6)</li>
-      <li><code>fillOpacity</code> (added in jQuery 1.6.2)</li>
-      <li><code>columnCount</code> (added in jQuery 1.9)</li>
-      <li><code>order</code> (added in jQuery 1.10.2)</li>
-      <li><code>flexGrow</code> (added in jQuery 1.11.1)</li>
-      <li><code>flexShrink</code> (added in jQuery 1.11.1)</li>
```
mgol, line 27: I removed it, but we could consider leaving the list in and updating the new values we added. But it's easy to just check what you have in your jQuery version and if you maintain a library, you shouldn't rely on this list anyway and just pass strings.

timmywil:
Do we want to document this list?

https://github.com/jquery/jquery/blob/447432f4a3b182cd6032930fea0685010e7a9d9c/test/unit/css.js#L1281

entries/css.xml
```diff
@@ -146,7 +147,15 @@ $( "div" ).on( "click", function() {
...
-      <p>When a number is passed as the value, jQuery will convert it to a string and add <code>px</code> to the end of that string. If the property requires units other than <code>px</code>, convert the value to a string and add the appropriate units before calling the method.</p>
+      <p>In jQuery 3.x or older, when a number is passed as the value, jQuery will convert it to a string and add <code>px</code> to the end of that string. There's one exception: <code>px</code> is not added to keys of <a href="/jQuery.cssNumber/"><code>jQuery.cssNumber</code></a> If the property requires units other than <code>px</code>, convert the value to a string and add the appropriate units before calling the method.</p>
```
timmywil, line 150: I'd like to change the wording here a bit. Each entry in `cssNumber` is an exception so there's technically multiple exceptions.

```suggestion
      <p>In jQuery 3.x or older, when a number is passed as the value, jQuery will convert it to a string and add <code>px</code> to the end of that string. However, there are exceptions. <code>px</code> is not added to keys of <a href="/jQuery.cssNumber/"><code>jQuery.cssNumber</code></a> If the property requires units other than <code>px</code>, convert the value to a string and add the appropriate units before calling the method.</p>
```

entries/jQuery.cssNumber.xml
mgol: As for the note, I added as slightly different now to align with what we have on other APIs like `.live()`.

timmywil: Discussed in meeting. The list is too outdated and we're just going to drop it.

mgol:
The list re-addded; here's how it looks right now:
Screenshot 2025-03-17 at 18 08 10

I wonder if it'd be better to group by version added instead?

PR #1264, opened 2024-10-19
jQuery: Add a self-closed tag back to an example

flinti:
This adds an actual quick-closed tag in this code example, as it is implied by the text that there can be one:

> Tags that cannot contain elements may be quick-closed or not:
```
$( "<img>" );
$( "<input>" );
```

to

> Tags that cannot contain elements may be quick-closed or not:
```
$( "<img>" );
$( "<input />" );
```


mgol (commented):
Thanks! This was the original intention of this code snippet and the `img` element was originally auto-closed; this was erroneously changed in 2f54b03ad716b8470d6f014865bc269e06c33ae1.

Maybe let's auto-close `img` instead of `input` again? This will be both closer to the past and would follow the order used in the preceding text.

flinti:
I just changed the PR to close `img` instead of `input`

mgol (approved):
Thanks!

PR #1272, opened 2025-04-07
Fixing keypress docs renaming wrong instances of "keyup"

### LINKED ISSUE #1266: keypress event mistakenly says "keyup"
https://api.jquery.com/keypress/

The event signatures both say "keyup" instead of "keypress":

> .on( "keyup" [, eventData ], handler )

> .trigger( "keyup" )

Seems to have been introduced in the last revision when .keypress() and similar methods were deprecated

aldimar-junior:
Fixes #1266


mgol (approved):
Oops! Thanks!

PR #1273, opened 2025-04-07
Changing wording from "cannot" to "should not"

### LINKED ISSUE #1244: Documentation error for text()
### Description ###
The jQuery documentation for `text()` at [https://api.jquery.com/text/](https://api.jquery.com/text/) states:

    The .text() method cannot be used on input elements. For input field text, use the .val() method.

But this is factually incorrect. jQuery allows a text value to be passed to an input, or any other empty tags such as `` for that matter. The provided content can instead be visible after the element on the console, as a child, although it will not show in the DOM.

For example, consider the following HTML chunk:

    <p><input /></p>

The following code

    $('input').text( "Hello" );

will render:

    <p><input />Hello</p>

Thus, the statement that "**.text() method cannot be used on input elements**" is factually incorrect, and maybe should be replaced with ".text() method **should** not be used on input elements"

Also, it should be added that the text thus provided will be added as a hidden child of the element.

aldimar-junior:
Changing the wording to indicate, as said by @dmethvin, that it's not supported and may not behave the same way in some future version.

Fixes #1244

### DIFF
--- a/entries/text.xml
+++ b/entries/text.xml
@@ -22,7 +22,7 @@
       <p>
         <code>Demonstration Box list item 1 list item 2</code>
       </p>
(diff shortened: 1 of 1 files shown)


mgol (approved):
Thanks!

Issue #1224, opened 2023-04-24
Document `jQuery.expr.pseudos`, `jQuery.expr[ ":" ]` & `jQuery.expr.filters`


mgol:
The main page should document `jQuery.expr.pseudos` but we should also have shorter pages for `jQuery.expr[ ":" ]` & `jQuery.expr.filters` as its deprecated aliases.

Sizzle docs can be a starting point as `jQuery.expr` is an alias of `Sizzle.selectors` in jQuery `<3.7.0`:
https://github.com/jquery/sizzle/wiki#sizzleselectorspseudosname--function-elem--

timmywil:
I don't know if it should. While the support probably can't go away, custom pseudos aren't necessarily something I'd want to encourage. It's better to use `.filter()` after selection.

mgol:
OK, let's discuss further at the team meeting. 🙂 cc @gibson042

timmywil:
We're going to change this issue to include the Sizzle wiki as its own API page, rather than document everything separately.

mgol:
PR: #1279

PR #1279, opened 2025-05-25
jQuery.find: Backport Sizzle wiki to an API page

mgol:
This change migrates most of https://github.com/jquery/sizzle/wiki to an API
page. Contrary to most other pages, each API didn't get its own XML `<entry>`
with a full signature to save time. This can be further improved in the future.

A Note section has been added, explaining the qSA to custom traversal fallback
mechanism of jQuery.

The browser support section has not been migrated as jQuery has different
browser support than Sizzle and it'd be irrelevant here.

Sizzle mentions have been replaced with `jQuery.find` or jQuery.

Fixes gh-1224

### DIFF
--- a/entries/jQuery.find.xml
+++ b/entries/jQuery.find.xml
@@ -1,0 +1,280 @@
+<?xml version="1.0"?>
+<entry>
+  <title>jQuery.find()</title>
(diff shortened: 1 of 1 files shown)


gibson042 (approved):
Nice!

mgol:
The page is live: https://api.jquery.com/jQuery.find/

PR #1298, opened 2026-03-10
Build: Skip checking code-of-conduct.openjsf.org & x.com in  the hydra crawler

mgol:
We're getting 429 Too Many Requests on almost every spider-check run.

We're already skipping twitter.com, let's do the same for x.com.

Example: https://github.com/jquery/api.jquery.com/actions/runs/22923413262/job/66527539538

### DIFF
--- a/.github/configs/hydra-config.json
+++ b/.github/configs/hydra-config.json
@@ -1,10 +1,14 @@
 {
   "//": [
     "2023-05: twitter.com serves broken redirect-loop",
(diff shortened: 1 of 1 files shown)


timmywil:
I see `twitter.com` in that list. We'll be adding x.com links soon, so how about we add `x.com` as well?

mgol:
@timmywil good callout! PR updated.

timmywil (approved):

mgol:
Success! https://github.com/jquery/api.jquery.com/actions/runs/22924380494/job/66530882236

